PR #7, opened 2021-03-01
get_codes() fails to process child JPN of R32JPN

francescolovat:
When calling `get_codes("node/R32")`, the region `R32JPN` has no child (`JPN`).

I've extended `test_nodes()` to check that, but it passes. So, it seems the problem is a bit more downstream. 

Will further investigate.

PS: I can't debug `get_codes()` since there is no module `sdmx`, but I think I've successfully pip install it (`sdmx-0.2.10`), so I'm not sure why it is still failing. Any suggestions @khaeru ?

### DIFF
--- a/message_ix_models/tests/model/test_structure.py
+++ b/message_ix_models/tests/model/test_structure.py
@@ -71,6 +71,7 @@ class TestGetCodes:
             ("R11", "R11_AFR", 50, "CIV"),
             ("R14", "R14_AFR", 51, "CIV"),
             ("R32", "R32SSA-L", 41, "CIV"),
+            ("R32", "R32JPN", 1, "JPN"),
         ],
     )
     def test_nodes(self, codelist, to_check, length, member):


khaeru:
> PS: I can't debug `get_codes()` since there is no module `sdmx`, but I think I've successfully pip install it (`sdmx-0.2.10`), so I'm not sure why it is still failing. Any suggestions @khaeru ?

setup.cfg for this package specifies a dependency of `sdmx1`:
https://github.com/iiasa/message-ix-models/blob/c3db1df34f107251a6aa7fc399cc57f19c560f9c/setup.cfg#L16-L23

Did you install the package? This should have installed the dependencies.

francescolovat:
> setup.cfg for this package specifies a dependency of `sdmx1`:

Yes, I did; installing `sdmx1`.

But now I see that in [here](https://github.com/iiasa/message-ix-models/blob/c3db1df34f107251a6aa7fc399cc57f19c560f9c/message_ix_models/model/structure.py#L6) there is an import of `sdmx` alone (and not `sdmx1`), that's why my editor does not recognize that module. Should I just change it that import statement to `sdmx1`?

francescolovat:
OK it was because I was not running this repo with my virtual env, where `sdmx1` was indeed installed.

khaeru:
> But now I see that in [here](https://github.com/iiasa/message-ix-models/blob/c3db1df34f107251a6aa7fc399cc57f19c560f9c/message_ix_models/model/structure.py#L6) there is an import of `sdmx` alone (and not `sdmx1`), that's why my editor does not recognize that module. Should I just change it that import statement to `sdmx1`?

No—like JPype, this is a work-around for an older, unmaintained package existing on PyPI: the name on PyPI and thus in setup.cfg or requirements.txt is `JPype1`, but code should `import jpype` (without the "1").

To back up a bit, what does "I can't debug" mean? Is it that an editor (e.g. PyCharm) fails to handle this situation, or the code actually fails to run with some specific error message?

francescolovat:
It works now as expected after setting the interpreter of this repo to the venv I used for iiasa/message_data#179.

I am not sure what was exactly happening, but after setting the right interpreter `get_codes()`, not it is correctly working in iiasa/message_data#179 as well.

PR #16, opened 2021-05-12
Add testing.export_test_data()

khaeru:
This generalizes a script provided as a Jupyter notebook by @OFR-IIASA in iiasa/message_data#216.

**To complete**
- [x] Add tests
- [x] Add to doc/whatsnew.rst

**To review**
- See the additions to the documentation.
- Try to run the example command.
- Check that the output is as expected.

### DIFF
--- a/message_ix_models/testing.py
+++ b/message_ix_models/testing.py
@@ -312,15 +312,15 @@ def export_test_data(context: Context):
 
     for name in reader.sheet_names:
         # Check if this sheet is to be included
-        if any(i in name for i in (EXPORT_OMIT + context.get("export_exclude", []))):
+        if name == "ix_type_mapping":
+            # Already handled
+            continue
+        elif any(i in name for i in (EXPORT_OMIT + context.get("export_exclude", []))):
             log.info(f"Discard sheet '{name}'")
 
             # Remove from the mapping
             ix_type_mapping.drop(name, inplace=True)
 
-            continue
-        elif name == "ix_type_mapping":
-            # Already handled
             continue
 
         # Copy the sheet from temporary to final file
--- a/message_ix_models/testing.py
+++ b/message_ix_models/testing.py
@@ -10,7 +10,7 @@ from ixmp import Platform
 from ixmp import config as ixmp_config
 
 from message_ix_models import cli
-from message_ix_models.util._logging import preserve_log_level
+from message_ix_models.util._logging import mark_time, preserve_log_level
 from message_ix_models.util.context import Context
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


khaeru:
@OFR-IIASA FYI, partway through this I retained the following code:
https://github.com/iiasa/message-ix-models/blob/4bf54e386bb6a6287a016d5cf4439c2c979ee517/message_ix_models/testing.py#L307-L318

The message "Filtering didn't work" is never printed. So I conclude the expanded list of filters to the initial `to_excel()` call achieves the same result. The next commit removes those lines.

khaeru:
@OFR-IIASA the PR checklist here is now complete, so this is ready for review as soon as you have a moment.

OFR-IIASA:
@khaeru in principal the function works well. I found that at a later point in time of creating the original input data, that I also excluded other sheets. Would it be useful therefore to also allow the OMIT list to be extended? I also included 
```
['ref_*', 'resource_*', 'extraction', 'commodity_stock', 'emission_scaling', 'relation_lower', 'relation_upper']
```
or shall I extend the list of sheets to be omitted?

khaeru:
> I found that at a later stage of creating the original input data, that I also excluded other sheets. Would it be useful therefore to also allow the OMIT list to be extended?

Hm, I guess it depends on _what_ particular test the data is needed for. Perhaps sometimes e.g. relation_upper might be wanted, so automatically deleting it is not ideal.

I will add a command-line option like `--omit=emission_scaling,relation_lower` that the user can provide. Then the docstrings for individual tests can describe which option values to give in order to prepare the data.

khaeru:
> I will add a command-line option like `--omit=emission_scaling,relation_lower` that the user can provide.

Okay, added, and tests pass. @OFR-IIASA please approve if there's nothing else to be desired at the moment.

message_ix_models/testing.py
```diff
@@ -201,3 +222,120 @@ def bare_res(request, context: Context, solved: bool = False) -> message_ix.Scen
...
+    "kpvs",
+    "lakl",
+    "land",
+    "lotol",
+    "mapping_macro_sector",
+    "MERtoPPP",
+    "prfconst",
+    "price_MESSAGE",
+    "sector",
+]
```
OFR-IIASA, line 249: Thanks for adding the option to add additional items to exclude. As for the hard-coded exclusions, I would request ref_activity,ref_extraction,ref_new_capacity,ref_relation to be added, as this is only very seldomly needed.

OFR-IIASA (changes requested):
One last adjustment regarding the exclusions; there ref_* parameter is hardly ever used hence this can also be hardcoded. I have tested the `exclude` function and everything is working well.

message_ix_models/testing.py
khaeru: Done!

message_ix_models/testing.py
```diff
@@ -207,3 +222,121 @@ def bare_res(request, context: Context, solved: bool = False) -> message_ix.Scen
...
+
+    # Load the scenario to be exported
+    scen = context.get_scenario()
+
+    # Retrieve the context settings giving the nodes and technologies to export
+    nodes = context.get("export_nodes", ["R11_AFR", "R11_CPA"])
+    technology = context.get("export_techs", ["coal_ppl"])
+
+    # Construct the destination file name
+    dest_file = private_data_path(
```
OFR-IIASA, line 278: Further testing of the command showed that a useful feature may be to allow a suffix to be added to the file output. This avoids files being overwritten accidentally.

OFR-IIASA (changes requested):

message_ix_models/testing.py
khaeru: I think this is somewhat overkill. This is a tool for development. It's on the individual developer to do things like:

- think about the test for which data is being exported,
- use appropriate options,
- choose names and locations that make clear what the data is and how it related to certain tests,
- handle files in their git working directory.

The code can't possibly anticipate all that, so I think it's best if it doesn't pretend to try.

OFR-IIASA (approved):

khaeru:
The codecov/project check failure is spurious: the lines _are_ covered when the tests are run locally, so it seems to be some issue with Codecov updating the check. Thus, not an obstacle to merging.

Issue #19, opened 2021-06-25
`List B` which lists the years in the ENGAGE scenario are incorrect.


OFR-IIASA:
The file https://github.com/iiasa/message-ix-models/blob/main/message_ix_models/data/year/B.yaml is incomplete. 2055 as well as 5 year timesteps prior to 1990 are missing.

PR #20, opened 2021-06-25
Year listing updated to match that of ENGAGE scenarios

OFR-IIASA:
The years listed in the `yaml` file which represent the years used in the ENGAGE scenario set, has been update to include:
a. 5 year timesteps prior to 1990.
b. 2055 added within the modelling horizon.


khaeru:
The GAMS license is stored as an encrypted secret, which means it's not available on a PR from a fork. See https://docs.github.com/en/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow

Can you please:
- Push to a new branch in message-ix-models, rather than in your fork?
- Take a moment to re-read the [code style docs per commit message format](https://docs.messageix.org/en/stable/contributing.html#code-style)?

khaeru:
Also please:
- When you open a PR that will close an issue (in this case #19), link it explicitly, so that merging the PR will close the issue.

OFR-IIASA:
Will be replaced by a new PR as suggested above.

PR #30, opened 2021-09-14
Updates to create compatibility for the Israeli model

ShaiWinograd:
**Updating and adjusting an existing code so that it will also work with "ISR" model**


khaeru:
This was rolled into #76.

PR #48, opened 2022-02-24
Modify the region name in R12.yaml

guofei2016:
Change the region name from R12_CPA to R12_RCPA in the R12.yaml file

In the development of the R12 model, in the very beginning we seperated the R11_CPA region inot two new regions, R12_CHN and R12_CPA. We finnaly decided to use R12_RCPA instead of R12_CPA as the new region name, so the region name R12_CPA in the R12.yaml file is outdated and it should be replaced by R12_RCPA. 

## How to review

please check the region name in the R12.yaml file 

## PR checklist

- [ ] Check the R12.yaml file and do some test if needed ✅


guofei2016:
@Jihoon @khaeru Thanks to both of you! This change should be done by a PR and I am so sorry for my previous mistake.

khaeru:
Replaced with #49 using a branch in Fei's fork.

PR #57, opened 2022-03-10
Add data descriptions for materials & transport

khaeru:
This adds a file that describes (in the EDITS format) the data requirements for an integration of MESSAGEix-Materials and MESSAGEix-Transport.

## How to review

TBA.

## PR checklist

- [ ] ~Add or expand tests;~ coverage checks both ✅
- [ ] Update doc/whatsnew.

### DIFF
--- a/message_ix_models/data/edits/transport-materials.yaml
+++ b/message_ix_models/data/edits/transport-materials.yaml
@@ -1,0 +1,22 @@
+format: 1
+
+title: Materials + transport representation in MESSAGEix
+description: |-
+  These are the data required to parametrize a representation of materials use
+  in the transport sector, i.e. the integration of MESSAGEix-Materials and
+  MESSAGEix-Transport.
+
+provider:
+  organisation: International Institute for Applied Systems Analysis (IIASA)
+  contact: xum@iiasa.ac.at; lorenzo.usai@ntnu.no; kishimot@iiasa.ac.at
+
+classifiers:
+- "Kind :: Model input"
+
+dimension:
+  time: {}
+  space: {}
+
+measure: {}
+
+quantity: []


khaeru:
@lorenzousai @THUminghao if you look at the other files in the same directory, you can see an example of the structure we should aim for here, e.g. quantities, measures, dimensions (with scope and resolution), etc. Please get in touch via Slack/Teams if you are not sure how to express the particular data needs/inputs in this format.

khaeru:
@lorenzousai @THUminghao thanks for each adding files here. Inspecting these allows us to make a many useful observations to guide the search for data; to know where data can be useful to both; or where assumptions must cover for data that's not available.

- There are some **similar concepts**. For example, LU uses `powertrain_type` while XM uses `powertrain_technology`; these are the same concept. LU's file notes that only certain values (BEV, PHEV) are required for his research, while XM's needs a larger set.
  - For these it's best to decide on a common name.
- The **`time` scope** has some *overlap* but also differences:
  - XM would like historical data to 1990; for LU, to 2010 is sufficient.
  - Both would need measurements as close to the present as possible.
  - From 2022, the data needed are *projections*/estimates/assumptions, up to 2100 in both cases.
- LU's file specifies 4 measures (Li_content, Co_content, etc.)—but these imply another dimension, the same as XM's `material_type`. So this suggests a simplification: those 4 measures (and corresponding 4 quantities) could be collapsed to 1 (and 1) with a similiar concept/dimension.
- Both are interested in concepts around the size/type of vehicles. We know this can be captured in different ways, e.g.:
  - XM has `vehicle_type` (general) and `make_model` (specific);
  - LU has `vehicle_segment`.

  In both cases, the idea is that different vehicles require different components of different size.
- LU's analysis requires data with detail on the `battery_technology` concept/dimension, whereas XM's (implictly) can use aggregates (averages/totals) across the mix of different technologies in-use/projected mix of different technologies.

Can I ask you to correspond with one another (here, on Slack or Teams, etc.) and both make edits to bring your files further into alignment?

FYI @vruijven @volker-krey, an outcome of the transport meeting on 28 April.

PR #60, opened 2022-03-29
Add a generic Workflow class

khaeru:
This is a straw-man PR to stimulate discussion, hence draft status. It adds a `Workflow` class to manage multi-scenario workflows.

## Requirements
The common pattern/apparent requirements for workflows is:

- Each scenario has zero or 1 (or more?) base/precursor/antecedent scenarios. These must exist before the target scenario can be created.
- The scenario is cloned from the precursor scenario.
- Modifications are applied. These can be modifications to **structure** or to **data**. For example:
  - Adding the MESSAGEix-Materials structure to a base MESSAGEix-GLOBIOM model.
  - Changing policy variables via constraint parameters.
  - Any other possible modification.
- The resulting scenario is maybe solved.
- The resulting scenario is reported.

Some workflows involve many such steps.

## Implementation

The implementation is based on the observation that these form a graph (DAG) of nodes (scenarios) and edges (steps), in the same way that reporting calculations do; and so the `dask` DAG features (via `genno`) can be used to organize the workflow.

The idea is the class could be used to remove all the common/duplicated code in the `runscript{,_main}` and `scenario_runner` modules in multiple places in `message_data`. These would be condensed to collections of small, simple, reusable functions, and then one function that bundles them into a complete Workflow.

## How to review

- Look at the added test_workflow.py.
- Comment on missing requirements (i.e. features of the current code), *especially* those that do not obviously fit into this implementation/pattern.

## PR checklist

- If this will proceed, implement additional required features, i.e. convert the following into checkboxes:
  - `run_reporting_only`
  - Allow additional/generic options to individual workflow steps.
  - Load a workflow from a file, probably YAML.
- [ ] Update doc/whatsnew.


gidden:
As I begin to look at the proposed implementation (thank you @khaeru!), let me start by describing my current experience.

I would break down my interaction with the framework in two phases: exploration (i.e., scenario creation) and production (running large-scale scenario ensembles).

At the end of my exploration phase, I will have a (perhaps large) set of changes that I need to employ to either create a one-time baseline scenario for my study or to apply to that baseline to create sub-experiment versions (e.g., a 'no novel technologies' case).

In the 'production' phase, I have defined the dimensions along which I want to make more atomic changes (e.g., 'high' vs 'low' biomass constraints). I then need to systematically sample this parameter space in order to create and solve scenarios of interest.

These production runs may require multiple input scenarios. From my experience I have seen at most 3 scenario dependencies

1. a 'parent' (one of the above 'baseline's)
2. (maybe) a previous scenario from which I take information from it's result (e.g., limiting activities or emissions based on the solution)
3. (maybe) a (different) previous scenario from which I take other information (e.g., end-use service demands from a deep-mitigation scenario)

In the 'production' phase, there are also two sub-phases, model generation/creation and solving. At the moment, the 'solving' step is actually overloaded, as it takes a scenario with a given structure (e.g., sampled from the parameter space described above) and applies a specific mitigation target (so, e.g., a generated model can be solved with a 500 budget, a 1000 budget, and so on).

Finally the 'solve' step is further overloaded. There are two 'flavors' of solves at the moment, the so-called 'peak' and 'full-century'. The 'full century' version places a single cumulative emission constraint on a scenario. The 'peak' scenario consists of a sub-workflow in 3 steps implementing the [new scenario logic](https://www.nature.com/articles/s41586-019-1541-4):

1. a cumulative co2 constraint with a floor constraint (currently 0, but could be different)
2. take solution from 1 and apply it directly as a period-by-period co2 constraint (to calculate carbon prices)
3. take solution from 2 and apply carbon prices also to other GHGs

I think as is clear from this description, there is ample opportunity to standardize aspects of this process and create reusable components as suggested here.

khaeru:
Thanks for these notes from your experience. Indeed, they all seem to fit into a workflow pattern.

I did consider the possibility of 2 or more precursor scenarios, but wanted to get more information on real use-cases. I think the code can be pretty easily extended to handle these.

khaeru:
> At the end of my exploration phase, I will have a (perhaps large) set of changes that I need to employ to either create a one-time baseline scenario for my study or to apply to that baseline to create sub-experiment versions (e.g., a 'no novel technologies' case).

Another thought here: for actual reproducibility, it is still necessary to package this set of changes as something like `create_base_scenario_for_this_project()`, and specify to which existing scenario it should be applied.

For one project, the code that sets up the workflow can flexibly (according to need/CLI option/config) either include that step or (e.g. if it is costly/time-consuming) instead load an already-created base scenario.

khaeru:
This closes #63 as noticed by @OFR-IIASA.

PR #83, opened 2022-10-13
Add Zambia yaml file and function same_time, analogous of same_node

adrivinca:
This PR makes to small additions:

- Adding a Zambia yaml file, needed to run the national model for different users in the LEAP-RE project
- Adding a new utility function `same_time` which does copy the sets `time` into either `time_origin` or `time_dest`. This function is the analogous of `same_node`

## How to review

There is not much to review or test. But please let me know if you notice something missing or needed

## PR checklist

- [x] Updated the automatic documentation
- [ ] Update doc/whatsnew. I did not, not sure if those changes are worth being mentioned there

### DIFF
--- a/message_ix_models/data/node/ZMB.yaml
+++ b/message_ix_models/data/node/ZMB.yaml
@@ -1,9 +1,10 @@
-# Codes for the "node" dimension of the MESSAGE-IL model
+# Codes for the "node" dimension of the MESSAGE-ZM model
 
 World:
   name: World
-  description: MESSAGE-ZA regions
+  description: MESSAGE-ZM regions
 
 Zambia:
   name: Zambia
   parent: World
+  child: [ZMB]
--- a/message_ix_models/data/node/ZMB.yaml
+++ b/message_ix_models/data/node/ZMB.yaml
@@ -1,0 +1,9 @@
+# Codes for the "node" dimension of the MESSAGE-IL model
+
+World:
+  name: World
+  description: MESSAGE-ZA regions
+
+Zambia:
+  name: Zambia
+  parent: World
--- a/doc/api/util.rst
+++ b/doc/api/util.rst
@@ -45,6 +45,7 @@ Commonly used:
    package_data_path
    private_data_path
    same_node
+   same_time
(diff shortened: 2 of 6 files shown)


message_ix_models/data/node/ZMB.yaml
```diff
@@ -0,0 +1,9 @@
+# Codes for the "node" dimension of the MESSAGE-IL model
+
+World:
+  name: World
+  description: MESSAGE-ZA regions
+
+Zambia:
+  name: Zambia
+  parent: World
```
khaeru, line 9: We found with MESSAGEix-Israel that it was important to have the `child:` field here with the ISO 3166-1 alpha-3 code for Israel:

https://github.com/iiasa/message-ix-models/blob/be551575085beb0d223c27e0aac5bbd392a6babf/message_ix_models/data/node/ISR.yaml#L7-L10

Can you please add that?

Among other things, this will allow to use this list with other code that selects/aggregates data from data sources that are at country resolution.

message_ix_models/data/node/ZMB.yaml
```diff
@@ -0,0 +1,9 @@
+# Codes for the "node" dimension of the MESSAGE-IL model
+
+World:
+  name: World
+  description: MESSAGE-ZA regions
```
khaeru, line 5: I would strongly suggest to use either “MESSAGE[ix]-Zambia” or “MESSAGE[ix]-ZM,” because [ZA is the ISO 3166-1 alpha-2 code for South Africa—**not** Zambia](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#ZA).

khaeru (changes requested):
Thanks for these improvements! I made a couple of small requests inline. Please also update doc/whatsnew—you can use the lines describing the addition of the ISR node codelist and of `same_node()` as examples.

adrivinca:
Thanks a lot for the quick check!
I addressed the points and also noticed another documentation file (`node.srt`) where I added ZMB

khaeru (approved):

PR #86, opened 2023-01-24
Add MESSAGEix-Nexus module

awais307:
This PR migrates MESSAGEix-Nexus module from `message_data` preserving the commit history. 

## How to review
TBD

## PR checklist

- [ ] Add, expand, or update documentation where required
- [ ] Update doc/whatsnew.
- [ ] Deciding on releasing the data files either through Zenodo, Git LFS or both 
- [ ] Agree on the directory structure so other modules can follow similar pattern
- [ ] Testing the scripts and adjusting where required 

## How to migrate

You will need to install [git-filter-repo](https://htmlpreview.github.io/?https://github.com/newren/git-filter-repo/blob/docs/html/git-filter-repo.html), e.g. by running `pip install git-filter-repo`, but see also [the official instructions](https://github.com/newren/git-filter-repo#how-do-i-install-it) and [their extensions and known problems](https://github.com/newren/git-filter-repo/blob/main/INSTALL.md). Furthermore, create a new folder and fill it with fresh clones of message_data and message-ix-models. 
Then, begin by navigating to the directory of message_data and checkout your feature branch:
```
git checkout nexus_move
# only keep our folders of interest
git filter-repo --path data/water/ --path message_data/model/water/ --force
```
Here, `water` is kept as an example to show the different paths where the new feature files will usually reside. Please exchange with your own paths. Also note that `--force` should never be used lightly as this could destroy your git history. However, starting from fresh clones provides a security net and checking out a branch requires `git filter-repo` to be run with `--force` initially. Next, we rename the paths to fit the message-ix-models naming scheme:
```
git filter-repo --path-rename data/water/:message_ix_models/data/water/
git filter-repo --path-rename message_data/model/water/:message_ix_models/model/water/
```
Afterwards, move to the message-ix-models directory (`cd ../message-ix-models/` if both are installed in the same parent directory) and add our freshly filtered message_data as a remote source:
`git remote add message_data ../message_data`
Update the remote source and create and switch to the branch of message-ix-models that is going to contain the new feature:
```
git remote update
git checkout -b migrate_nexus
```
Finally, we can merge the branch to message-ix-models:
`git merge --allow-unrelated-histories message_data/nexus_move`
To push this branch to message-ix-models, you might be able to run
`git push --set-upstream origin migrate_nexus`

If your branch uses or had used git lfs, you might run into additional problems trying to push to message-ix-models of the sort
`error: failed to push some refs to 'github.com:iiasa/message-ix-models.git'`. To remedy those, you can follow these instructions:
First, collect all the lfs files git can still find on their servers:
```
git lfs fetch --all
git lfs ls-files --all -l | \
  perl -n -e 'chomp; @x = split; next if $x[1] eq "*"; $o = $x[0]; $p = substr($o, 0, 2) . "/" . substr($o, 2, 2) . "/$o"; print "$o\n" unless -f ".git/lfs/objects/$p";' \
  >/path/to/lfs-sha256s-to-remove.txt
```
This collects a list of sha256s in the specified location. With these, we can collect the corresponding sha1s, which `git filter-repo` can then use to filter these objects from our history:
```
git rev-list --all --objects -- | \
  cut -d' ' -f1 | \
  while read sha1; do
    if git cat-file -p "$sha1" | grep -q -F -f /path/to/lfs-sha256s-to-remove.txt; then
      echo "$sha1";
    fi;
  done \
  >/path/to/git-sha1s-to-remove.txt

# remove from history
git filter-repo --strip-blobs-with-ids /path/to/git-sha1s-to-remove.txt --refs migrate_nexus --force
```
Note that these commands are bash commands and may not work out of the box for alternatives like fish. Make sure to only remove these objects from the history of our new branch (using the `--refs` option). 

Lastly, define the upstream of our new branch to be the public repo and push the branch there:
`git push --set-upstream origin migrate_nexus`

This might, however, lead to unrelated histories that don't want to let us merge. One way around this (tested and worked in this case) is by creating another branch from main:
```
git checkout main
git checkout -b nexus_move
git merge migrate_nexus --allow-unrelated-histories
git push -u origin nexus_move
```
It might be more elegant (but remains to be tested) to do the following alternatively: Before leaving message_data, but after running all filter commands there, save it's initial commit in a variable:
```
FEATURE_BASE=$(git rev-list --max-parents=0 HEAD)
[ -n "$FEATURE_BASE" ] || exit 1
```
The second line makes sure the first command was successful and exits the run if it wasn't. Then, on the main branch of message-ix-models (so before checking out our new branch there), save its initial commit in a similar way:
```
BASE=$(git rev-list --max-parents=0 HEAD)
[ -n "$BASE" ] || exit 1
```
Now, we can replace the initial commit of our branch with the initial commit of the main branch, uniting their commit histories (and cleaning up with `git filter-repo` afterwards):
```
git replace --graft $FEATURE_BASE $BASE
git filter-repo --force
```
This way has not been tested, though, and might cause problems. In particular, you might have to run the `FEATURE_BASE` part after merging the message_data branch to the message-ix-models feature branch as the commit hash might change in the process. If all is fine, you should also be able to run
`git push --set-upstream origin migrate_nexus`
and merge to the main branch after creating a PR.


glatterf42:
Talking with @khaeru, we found a few things that need to be cleaned up before this PR can be merged. For some of them we might be able to use git-filter-repo again, but I have some questions first:

- In `message_ix_models/model/water/data`, there is a directory called `pre-processing`. This naming won't allow python files from within it to be imported anywhere else (this also holds for e-flow.py within that directory). At first glance, they don't seem to be imported anywhere, but that desire might arise in the future, so it should probably be renamed (e.g. to pre_processing and e_flow.py)
- In the very same directory, there is a folder called `deprecated`. Does this need to be in the PR?
- In `message_ix_models/data/water/demands/harmonized/`, there is a folder called `old_R11`. This one also doesn't seem to be referenced anywhere, so could we delete it altogether? 
- In `message_ix_models/data/water/`, there is a file called README.txt. Is this specific to this folder? Otherwise or even in that case, we might want to move it to the doc/ folder of message-ix-models. This file also contains a reference to the P drive so that I don't know if we'd want to publish it.

message_ix_models/data/water/README.txt
```diff
@@ -0,0 +1,9 @@
+The data and scripts used to generate the shapefiles are located in the interla IIASA drives
```
khaeru, line 1: Text from this file should be moved to the documentation pages.

khaeru:
@glatterf42 could you please update the PR description here with the full set of commands used to derive this branch from `message_data`? This will help us iterate on it until it gives the changes we need.

khaeru:
> @glatterf42 could you please update the PR description here with the full set of commands used to derive this branch from `message_data`?

I finally found that old script I was thinking of: https://github.com/khaeru/genno/blob/385ef350c0ef5a3a9ac3ee88b6fbc83854088f6d/bootstrap.sh
(This was used when extracting the `genno` package core from `ixmp.reporting`.)

Please look particularly at the last few operations ("Store the initial commit of OTHER", "Graft the $OTHER/[main] branch onto the first commit of [main] so the two share a common initial commit")—I think these are likely what we are missing here to get a linear history that git and GitHub both can understand.

glatterf42:
Wonderful, thank you, @khaeru. I didn't have time to test your new method yet, but iirc our other way also produced a linear commit history? Either way, both are noted in the updated PR description.

khaeru:
Closed as superseded by #88.

PR #88, opened 2023-02-15
Add MESSAGEix-Nexus module

khaeru:
This PR migrates `.model.water` (i.e. MESSAGEix-Nexus) from the (private) `message_data` repo to this `message-ix-models` package, following the procedure described in #89.

Supersedes #86.

## How to review

Note the CI checks all pass, with the exception of the coverage checks, per notes below.

## PR checklist

- [x] Complete and merge #91 into this branch.
- ~Add or expand tests; coverage checks both ✅~ See https://github.com/iiasa/message-ix-models/pull/88#issuecomment-1443375799; will be handled via #93.
- [x] Update doc/whatsnew.


adrivinca:
I create [migrate-nexus-adj](https://github.com/iiasa/message-ix-models/tree/migrate-nexus-adj) to:

- add some data files that were missing after the migration
- rename `message_ix_models` instead of `message_data`
- change the name of the cli `water` function
- fix lint issues
anything else that should be done there?

However, there will be some functions that depend on other parts of message_data (afaik only the following) that will not work. I am listing them here and will add some warnings in the documentation

`from message_data.tools.post_processing.iamc_report_hackathon import (
    report as old_reporting,
)`

khaeru:
> However, there will be some functions that depend on other parts of message_data (afaik only the following) that will not work. I am listing them here and will add some warnings in the documentation
> 
> `from message_data.tools.post_processing.iamc_report_hackathon import ( report as old_reporting, )`

For this, we can use this kind of pattern:
```python
try:
    from message_data.tools.post_processing.iamc_report_hackathon import (
        report as old_reporting
    )
except ImportError:  # message_data not installed
    old_reporting = None
```

or:
```python
from message_ix_models.util import MESSAGE_DATA_PATH

if MESSAGE_DATA_PATH is None:
    def old_reporting(*args, **kwargs):
        pass
else:
    from message_data.tools.post_processing.iamc_report_hackathon import (
        report as old_reporting
    )
```

These patterns are common in other packages that have optional dependency; e.g. package `foo` is installed, and it should work if the user also installs package `bar`, but also if `bar` is not installed.

The idea is:
- When `message_data` is available, then the correct function/other object is imported from the correct module.
- When `message_data` is not available, then the above code can be _imported_ without any error. (This is helpful in case, e.g. setting up the CLI involves importing code from a `.model.water` module that in turn imports from `message_data`.) But trying to _run_ it would produce an error/unexpected results, because the objects (e.g. `old_reporting`) are just mock-ups/`None`.

glatterf42:
It looks like not a single test has been written for the files here, which I find hard to believe. Also, I thought we excluded the files from model/water from test coverage, but maybe that was on the old branch? Would we want to exclude them from tests here or add tests for them (by writing them or importing the ones that got lost)?

khaeru:
> It looks like not a single test has been written for the files here, which I find hard to believe. Also, I thought we excluded the files from model/water from test coverage, but maybe that was on the old branch? Would we want to exclude them from tests here or add tests for them (by writing them or importing the ones that got lost)?

Good catch. It looks like there _were_ a few tests in `message_data`: https://github.com/iiasa/message_data/tree/dev/message_data/tests/model/water.

Checking closely, neither the process described in #86 or #89 included this path. Let me make a commit to #89 to add them. Then:
- Someone should re-try the process to see that it preserves these files. (I can't do this right away, but could later today or next week.)
- This would update `migrate-nexus` via force-push.
- We could then rebase `migrate-nexus-adj2` as used in #91 onto the updated `migrate-nexus`.

These last two steps are why it was prudent to keep these two branches (automatic & manual) separate.

khaeru:
> Also, I thought we excluded the files from model/water from test coverage, but maybe that was on the old branch? Would we want to exclude them from tests here or add tests for them (by writing them or importing the ones that got lost)?

- Even once we migrate the tests (per the prior comment), coverage will be less than 100%.
- I think there is a difference here:
  - For development code in `message_data`, it's okay to silence warnings about lack of coverage. In some cases moving quickly there is a higher priority than having robust code (via testing).
  - For `message-ix-models`, robustness is more important. When we don't have the resources (funding, people's time) to write tests, then openly acknowledging that some code is untested is an honest and helpful signal to would-be users: "be careful, this is not tested; it may not behave as advertised or work in your intended use-case." With that info, users can still use the code, but know when to exercise caution.
- Thus, I would prefer we do not exclude files from coverage in `message-ix-models`.
- However, we don't want to drag out merging the current PR, either.
- Rather, we should just acknowledge:
  - Merging this PR will lower test coverage from 100% to [new value] (seems 60–70%).
  - We create a follow-up issue like "Improve test coverage of `.model.water`".
  - We advocate that people get time to create 1 or more PRs to address that issue.

adrivinca:
we agree on merging it even if it reduces the coverage

PR #94, opened 2023-03-07
Satisfy mypy complaints due to nexus merge

glatterf42:
Fix the errors mypy throws due to the recent merger of the nexus/water module.

These errors seem to fall into two broad categories: the classes `Annotation` and `Code` from sdmx1 require `InternationalString`s to be passed as text arguments. I have not yet worked with them, so I don't know for sure that I kept the intended behaviour. Secondly, several functions from the dataclasses module require arguments to be a `DataclassInstance` or of `Type[DataclassInstance]` or both. Curiously, the dataclasses-provided function`is_dataclass` doesn't just return True when dealing with a `DataclassInstance`, but also when dealing with a `Dataclass` itself. So passing this function is not enough to ensure we're dealing with a `DataclassInstance`. Again, I've fixed the conditions so that the correct data types are provided (in util/config.py and util/common.py), but I don't know if I kept the intended behaviour. E.g. in [config.py](https://github.com/iiasa/message-ix-models/blob/d9287a8b95073ecf6b8280027393bc5b030a09b1/message_ix_models/util/config.py#L91-L98), the `replace` function is the first one to actually require existing to be a `DataclassInstance`, but maybe that check would also make sense before, i.e. when `is_dataclass` is called. 

## How to review

Please check that I kept the intended behaviour. 

## PR checklist

- ~[ ] Add or expand tests; coverage checks both~ Only type-hint related changes intended
- ~[ ] Add, expand, or update documentation.~ Only type-hint related changes intended
- ~[ ] Update doc/whatsnew.~ Only type-hint related changes intended

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -135,3 +135,6 @@ dmypy.json
 # Generated and temporary data files
 debug/
 cache/
+
+# VSCode settings
+.vscode
--- a/message_ix_models/model/disutility.py
+++ b/message_ix_models/model/disutility.py
@@ -6,7 +6,7 @@ from typing import List, Mapping, Sequence, Union
 
 import message_ix
 import pandas as pd
-from sdmx.model import Annotation, Code
+from sdmx.model import Annotation, Code, InternationalString
 
 from message_ix_models import ScenarioInfo, Spec
 from message_ix_models.model.build import apply_spec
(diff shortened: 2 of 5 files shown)


khaeru:
Per the InternationalString errors, these are upstream. Can you please check with the branch from khaeru/sdmx#117? If that causes them to disappear, we can avoid the respective changes here.

khaeru (approved):
Thanks for the fixes!

PR #99, opened 2023-05-11
Add technoeconomic cost projections and related data

measrainsey:
This PR adds a module that projects regionally-differentiated investment costs and fixed O&M costs for technologies.

The module projects costs in a specified reference region for all technologies by applying a decay/reduction rate. Then, costs for all other regions are calculated through one of the following methods:
1. Constant ratio: all regions' base year relative ratio to the reference region is applied through the projection period, so all regions follow the same reduction rate as the reference region
2. Convergence year: all regions' costs converge to the reference region by a specified year
3. GDP per capita-adjusted: projected GDP per capita from SSP scenarios are used to adjust the regional cost ratios into the future, using on the assumption that technology costs follow GDP per capita

Note that all costs calculations mentioned above are for investment costs. Fixed O&M costs are assumed to be a specified fraction of investment costs. 

The main function that projects costs is the `create_cost_projections()` function, which takes the following inputs:
- Node (R11, R12, or R20)
- Reference region (by default NAM for each node)
- Base year (by default, 2021 when possible, if not fall back to use 2020)
- SSP scenario version (updated/review or previous)
- Scenario (SSP1, SSP2, SSP3, SSP4, SSP5, LED)
- Projection method (learning, GDP adjusted, or convergence with splines)
- Convergence year
- Rate of fixed cost change over time
- Output format (MESSAGE, IAMC)

The `create_cost_projections()` outputs a class object with two dataframes (one for investment costs, one for fixed O&M costs).

## How to review

**Documentation and usability** (@adrivinca)

- Documentation has been added, under `General purpose modeling tools` > `Technoeconomic investment and fixed O&M costs projection`. The following link leads straight to the section in the latest docs test build: https://iiasa-energy-program-message-ix--99.com.readthedocs.build/projects/models/en/99/api/tools.html#technoeconomic-investment-and-fixed-o-m-costs-projection-tools-costs
- The `demo.py` script in the module gives some example of how to use the main function that creates the costs projections.
- Ensure that a colleague who has not used this tool before (a) understand and (b) use it, based on above materials provided.
  - Reviewer: read through, ask questions (as comments on specific lines, or in [this Slack thread](https://iiasa-ece.slack.com/archives/CD0GBHHA4/p1708594079641019)).
  - Author (@measrainsey): make edits to docs sufficient to preclude such questions.

**Technical, implementation** (@khaeru)

- Improve usage of the Config class.
- Other comments on the code.

## PR checklist

- [x] Update doc/whatsnew.


message_ix_models/tools/iea/weo.py
```diff
@@ -0,0 +1,672 @@
...
+"""
+
+import numpy as np
+import pandas as pd
+
+from message_ix_models.util import package_data_path
+
+
+def get_weo_data():
+    """
```
khaeru, line 12: For docstrings, per the [code style](https://docs.messageix.org/en/stable/contributing.html#code-style), we follow the numpydoc standard. See the links there to the standard, PEP 257, etc.

message_ix_models/tools/iea/weo.py
```diff
@@ -0,0 +1,672 @@
...
+
+    # Read in raw data file
+    file_path = package_data_path(
+        "iea", "WEO_2022_PG_Assumptions_STEPSandNZE_Scenario.xlsb"
+    )
+
+    # Dict of all of the technologies,
+    # their respective sheet in the Excel file,
+    # and the start row
+    tech_rows = {
```
khaeru, line 28: All of these fixed values can be moved outside functions—even if they're only used in one function.

As globals, they can be named in all caps, e.g. `TECH_ROWS`.

message_ix_models/tools/iea/weo.py
```diff
@@ -0,0 +1,672 @@
...
+        "marine": ["Renewables", 125],
+    }
+
+    # Specify cost types to read in and the required columns
+    cost_cols = {"capital_costs": "A,B:D", "annual_om_costs": "A,F:H"}
+
+    # Loop through each technology and cost type
+    # Read in data and convert to long format
+    dfs_cost = []
+    for tech_key in tech_rows:
```
khaeru, line 62: Use [`itertools.product()`](https://docs.python.org/3/library/itertools.html#itertools.product) instead of nesting loops.

message_ix_models/tools/iea/weo.py
```diff
@@ -0,0 +1,672 @@
...
+            df = pd.read_excel(
+                file_path,
+                sheet_name=tech_rows[tech_key][0],
+                header=None,
+                skiprows=tech_rows[tech_key][1],
+                nrows=8,
+                usecols=cost_cols[cost_key],
+            )
+
+            df.columns = ["region", "2021", "2030", "2050"]
```
khaeru, line 73: In general, try to use pandas method chaining.

This whole loop block could be collapsed to something like:
```
df = (
    excelfile.parse(…)
    .set_axis(["region", "2021, "2030", "2050"], axis=1)
    .melt(…)
    .assign(
       scenario="stated_policies",
       technology=tech_key,
       …
     )
     .reindex(["scenario", "technology", …], axis=1)
)
```
Especially using assign() can be much faster than separate indexed assignments.

message_ix_models/tools/iea/weo.py
```diff
@@ -0,0 +1,672 @@
...
+    """
+
+    # Replace "n.a." strings with NaNs
+    weo_df["value"] = weo_df["value"].replace("n.a.", np.nan)
+
+    # Filter for only United States data (this is the NAM region)
+    df_us = weo_df.loc[weo_df.region == "United States"].copy()
+
+    # Rename the `value` column in the US dataframe to `us_value`
+    df_us.rename(columns={"value": "us_value"}, inplace=True)
```
khaeru, line 137: Any time you notice yourself using `inplace=True`, especially successively, can be replaced with chaining.

message_ix_models/tools/iea/weo.py
```diff
@@ -0,0 +1,672 @@
...
+        & (df_cost_ratio.r11_region.isin(["EEU", "FSU"])),
+        "cost_ratio",
+    ] = 0
+
+    # Assumption 2: For pulverized coal with CCS and IGCC with CCS in MEA,
+    # make cost ratio the same as in the FSU region
+    # TODO: this method to replace the values seems a little prone to errors,
+    # so probably best to change later
+    df_cost_ratio.loc[
+        (df_cost_ratio.cost_ratio.isnull()) & (df_cost_ratio.r11_region == "MEA"),
```
khaeru, line 187: Use `fillna()`

message_ix_models/tools/iea/weo.py
```diff
@@ -0,0 +1,672 @@
...
+    "solar_i": "solarpv_buildings",
+    "heat_i": "ccgt_chp",
+    "geo_hpl": "geothermal",
+    "nuc_lc": "nuclear",
+    "nuc_hc": "nuclear",
+    "csp_sm1_ppl": "csp",
+    "csp_sm3_ppl": "csp",
+}
+
+conversion_2017_to_2005_usd = 83.416 / 103.015
```
khaeru, line 270: We could update this in the iam-units package used throughout, but not urgent.
Please state the source for these values.

message_ix_models/tools/iea/weo.py
```diff
@@ -0,0 +1,672 @@
...
+    return df_nam_assumptions
+
+
+# Type 1: WEO * conversion rate
+def adj_nam_cost_conversion(df_costs, conv_rate):
+    df_costs["cost_NAM_adjusted"] = df_costs["cost_NAM_weo_2021"] * conv_rate
+
+
+# Type 2: Same as NAM original MESSAGE
+tech_same_orig_message_inv = [
```
khaeru, line 347: Good to sort entries in these lists; it makes them easier for others to scan.

message_ix_models/tools/iea/weo.py
```diff
@@ -0,0 +1,672 @@
...
+        "cost_NAM_adjusted",
+    ] = df_costs.loc[
+        (df_costs.message_technology.isin(list_tech_fom))
+        & (df_costs.cost_type == "annual_om_costs"),
+        "cost_NAM_original_message",
+    ]
+
+
+# Type 3: Manually assigned values
+dict_manual_nam_costs_inv = {
```
khaeru, line 411: Again, state the source.

message_ix_models/tools/iea/weo.py
```diff
@@ -0,0 +1,672 @@
...
+        "reference_cost_type": "capital_costs",
+    },
+    "solar_i": {
+        "reference_tech": "solar_pv_ppl",
+        "reference_cost_type": "capital_costs",
+    },
+}
+
+dict_tech_ref_fom = {
+    "gas_ppl": {"reference_tech": "gas_cc", "reference_cost_type": "annual_om_costs"},
```
khaeru, line 542: Looks like "reference_" in the string keys could safely be dropped.
Use a comment where these are used to indicate these are references.

message_ix_models/data/costs/eric-fom-costs.csv
```diff
@@ -0,0 +1,70 @@
+# Eric's adjusted/manual fixed O&M costs for MESSAGE technologies
+#
+# Units: 2005 USD per kW
+# 
+# - This is copied directly from the RegionDiff sheet in SSP1_techinput.xlsx
```
khaeru, line 5: Please give the complete path to this file, whether on the IIASA P-drive or elsewhere, so that someone could locate it.

Rather than "Eric's" in the title and file name, perhaps use:
- fix_cost-0.csv as a file name—"fix_cost" aligns with the name of the MESSAGEix parameter that is eventually targeted; "-0" indicates there might be multiple alternatives (Eric's, Alice's, Bob's, etc.)
- Add a bullet like "These were prepared ca. [DATE] by Eric Lastname [details…]."

khaeru (commented):
This is a great start for a first message-ix-models PR!

I leave some comments about general pandas usage. I think with these applied the code will slim down considerably, making the intended behaviour more clear. That can then be targeted in a next review.

volker-krey:
It would be nice to include needed package dependencies into `message-ix-models` pip install command. I had to manually install `scikit-learn` manually for the code to run.

volker-krey:
In IAMC data template output, the Model and Unit columns are missing and should be added.

volker-krey:
Time steps (currently 10-year time steps) and set of regions for which to produce projections should become easily configurable. Also, region names used in GDP/cap specification, IEA WEO mapping, etc. then need to become unambiguous. I would suggest to use `RXX_NAME` (e.g., R11_NAM, R12_CHN) for global model regions.

khaeru:
Another thing to add:
- Among the checks listed above is the Read the Docs auto-build of the PR branch. Clicking that leads to https://iiasa-energy-program-message-ix--99.com.readthedocs.build/projects/models/en/99/ —note the subdomain includes the PR number.
- From there one can navigate to the added docs page: https://iiasa-energy-program-message-ix--99.com.readthedocs.build/projects/models/en/99/api/tools.html#technoeconomic-investment-and-fixed-o-m-costs-projection-tools-costs
- A direct link to this can be added to help reviewers.

pyproject.toml
```diff
@@ -44,6 +44,9 @@ dependencies = [
   "PyYAML",
   "sdmx1 >= 2.13.1",
   "tqdm",
+  "pyxlsb >= 1.0.10",
+  "scikit-learn",
+  "nomenclature-iamc",
```
khaeru, line 49: @measrainsey is `nomenclature-iamc` actually used in the code? I can't seem to find where.

message_ix_models/tools/costs/projections.py
```diff
@@ -0,0 +1,777 @@
...
+from message_ix_models.tools.costs.splines import apply_splines_to_convergence
+
+
+class projections:
+    def __init__(self, inv_cost, fix_cost):
+        self.inv_cost = inv_cost
+        self.fix_cost = fix_cost
+
+
+def smaller_than(sequence, value):
```
khaeru, line 28: According to coverage/pytest-cov/Codecov this function and `larger_than()` are not tested, and according to jedi/VS Code they are not used anywhere in the module. Safe to remove?

khaeru:
> 1. The use of the `learning` term which refers to percent cost reduction over time here. I thought some time ago in the MESSAGE meeting we decided to rethink the use of this term because it generally refers to cost changes as a function of deployment/R&D. Also, this might be confusing once the learning module, that Matt and I are developing, is integrated to MESSAGE.

What alternate terms were proposed? Can you link to that discussion?

I am also not sure the word "learning" is misapplied here. AFAICT the difference is not whether learning vs. something else is being modeled; it's rather the _model representation_ of real-world learning processes:
- In this module, if `method="learning"` is chosen, then deployment, R&D, and other real-world processes are _assumed to_ result in a certain rate of cost decrease over time, with the rate being _exogenous_, fixed, and configurable by the user.
- In the work you and Matt are doing, cost decreases are the result of an _endogenous_ representation of learning processes, with other (meta)parameters controlling this representation.

Am I missing something?

If not, then IMO the best approach is better documentation. In the future whenever that endogenous learning code lands in message-ix-models, both its docs and the docs for this module can be updated with notice blocks like "Do you want to model endogenous learning? Don't use this module, instead use [link]." and "Do you want to model learning with simpler, exogenous assumptions of learning rates? Instead use .tools.costs()". Users could be cautioned (or prevented) to avoid using both simultaneously, etc.

ywpratama:
> > Just for clarity in the docs, could we be more specific on the sign for fom_rate, e.g., positive fom_rate means cost increase, negative means decrease, or the opposite? does the default value of 0.025 mean that, by default, FOM cost is assumed to increase by 2.5% annually? (https://iiasa-energy-program-message-ix--99.com.readthedocs.build/projects/models/en/99/api/tools.html#message_ix_models.tools.costs.Config.fom_rate)
> 
> @ywpratama A positive `fom_rate` does mean fixed O&M costs will increase over time (for that specific technology-vintage). The actual equation used to calculate the fixed O&M cost in a certain year for a certain technology-vintage is:
> 
> ```
> x.fix_cost * (1 + (fom_rate)) ** (x.year_act - x.year_vtg)
> ```
> 
> Here, `fix_cost` is the fixed cost at year vintage = year actual. Now, whether I can remember _why_ that specific formula was used is a different story 😅

Ah, I see. In that case, I think that `fom_rate` has a positive value and calculated using the formula above is to show that the FOM increases as the plant is aging. We have this formula as well in daccs module.

adrivinca:
Hi, here is my review

### Documentation. 
now it shows the automatically generated documentation for all the functions. 

- I think an introduction paragraph written in human language is missing. What is the module doing, why, in which possibilities (energy, material), what are the current sources of data? and format outputs (message, iamc)
- a small example like one of those in `demo.py` should also be written in the first part of the documentation, so that the reader is immediately pointed to the config and `create_cost_projections` function. Otherwise, this info don't emerge clearly from the documentation

Concerning the labeling "learning", I think it is fine, although I was initially confused because here in the PR description you call the "learning" option "constant ratio".

### Concerning the package itself

ERROR
- due to `project/ssp/data.py` because it first checks if I have `message_data`, which I have but not on the right branch, and throws an error. I think all the references to message_data should be removed. [see slack thread](https://iiasa-ece.slack.com/archives/CD0GBHHA4/p1708963310171759)

data:
is the IEA data also used in other applications? or do we expect more IEA data to be added in the `data/iea` folder for other uses? maybe better to keep it under costs, so that there is always some clarity on who is in charge of updating it, or where it belongs. The risk is that more IEA data, related to other projects will be added to the folder, and an overview will be lost. But this can be discussed of course.

khaeru:
@measrainsey I am seeing a few of the CI jobs failing a single test:
> FAILED message_ix_models/tests/util/test_cache.py::test_cached - IndexError: list index out of range

I think this is probably due to #156 (which updated logging in this package, message-ix-models) or the release of genno 1.25 (which updated caching). I will try to investigate/diagnose/fix and will let you know when it does. In the meantime please proceed as if that test is passing.

khaeru:
> I will try to investigate/diagnose/fix and will let you know when it does. In the meantime please proceed as if that test is passing.

Could you please try to rebase? That may fix it.

measrainsey:
Would it be possible to get your eyes on this PR again please? 👀 @khaeru 

Major updates since last review was:
- I removed the use of splines completely (and `splines.py` and `test_splines.py` have been deleted).
- I renamed the `learning` projection method to `constant` (to reflect that the regional cost ratios are kept constant over the projection period).
- I updated the docs based to reflect both the changes above and the feedback from Yoga and Adriano to make the documentation more human-readable. 

The tests are currently failing. In addition to the CI jobs tests failing as mentioned by @khaeru above, I am also seeing the tests failing for the `exo_data` with the main error being `ValueError: No source found that can handle 'ICONICS:SSP(2024).4'`. I wonder if that is because we changed the code for `exo_data` in this PR but not any of the corresponding tests?

Let me know how you would prefer to move forward with this @khaeru, or if you think any additional work is needed. I think I'm done with adding things for the most part, so once everything looks good to you I am happy to merge. Thanks!

khaeru:
At first glance I see a couple of minor clean-ups:
- There is a commit with subject "TEMPORARY Pass test_apply_splines_to_convergence()" still on the branch. Since you've removed the code this applied to, I would suggest to rebase and drop the commit, which is our usual practice to avoid having these temporary commits in the history of the `main` branch.
- There is an unanswered question/comment on pyproject.toml.

I'll take a more detailed look early next week.

khaeru:
There's a conflict (I suspect one of the items we discussed) after the merge of #162. So I'll:

- Rebase, resolving the conflict.
- Address the TEMPORARY commit mentioned above.
- Complete "Update doc/whatsnew" from the PR checklist in the description.
- Review and approve.

khaeru (approved):
Okay, good to go.

There are some test failures, but I think #165 will fix those, so we'll rebase after that's in and—if all greens—merge.

Issue #93, opened 2023-03-06
Improve `.model.water`


khaeru:
This is a tracking issue for follow-up items from #88. Create 1 or more PRs and (if necessary to keep discussion organized) issues to address:

- Add tests to raise coverage to 100%.
- Consider methods for packaging data, e.g. omit some data from the built message-ix-models package, and provide a script to retrieve it from GitHub, Zenodo, or other location(s).
- Provide link(s) to the GMD manuscript in review or as published.
- Rename the CLI from `mix-models water-ix` to `mix-models water` (after corresponding change has been made on `message_data`).
- (Add items here.)

khaeru:
> Consider methods for packaging data, e.g. omit some data from the built message-ix-models package, and provide a script to retrieve it from GitHub, Zenodo, or other location(s).

Just for information here, I tried the following:
```
$ python setup.py bdist_wheel sdist
```
- With the added files in `message_ix_models/data/water/`, the size of the total (compressed) package is now 2.7 MB.
- Unpacked, those files are about 10.8 MB.

The fact that you've chosen to handle this data as CSVs rather than binary formats like MS Excel means that it compresses quite well. This is also not a lot compared to other packages (e.g. numpy is 10–20 MB compressed), so my sense is that it is not, for now, worth the effort to invent a more elaborate way of handling this data.

PR #106, opened 2023-06-19
Update the nexus module

awais307:
This PR addresses [issue#93](https://github.com/iiasa/message-ix-models/issues/93) and updates data to the nexus module, addresses the tests issue

The code has been restructured to be more test-friendly

two tests has been marked with `@pytest.mark.xfail(reason="Temporary, for #106")`, namely
`test_report_full` in test_report.py
and `test_build` in test_build.py
as they both require a fully parametrized scenario as a starting point, which is not the case for the "bare" scenario for tests.

Since we are about to do major revision and update to the module, we decided to postpone investing further time in those tests.

all tests seem to pass for py3.12, not for py3.11

## PR checklist

- [x] Update doc/whatsnew.


glatterf42:
Currently, the linting test is failing (among others). Please see our guide for [contributing code](https://docs.messageix.org/en/stable/contributing.html#code-style) to see what code style we expect. Please also note that black, flake8, mypy, and isort can all be configured to be applied automatically whenever you save your code (the details of which can vary, but we can work it out together if you want to). This helps passing theses tests tremendously.

glatterf42:
The current issue with the other tests seems to be that they still rely on code in message_data (which they shouldn't, as explained before, since all code here should be open source):

> message_ix_models/tests/model/water/test_nexus.py:5: in <module>
    from message_data.model.water import build
E   ModuleNotFoundError: No module named 'message_data'

glatterf42:
Also to make sure: how large are these data files that you are updating here? If they are too large (and there are lots of them), it might be better to upload them to Zenodo or so (see the second point in [the comment](https://github.com/iiasa/message-ix-models/issues/93#issue-1611612780) by @khaeru).

khaeru:
- Changed to "draft" status; prefer this over putting "[WIP] " in the title, as it means the same thing.
- I would recommend to update the branch by "Update with rebase" and then `git pull --rebase` on your local.

glatterf42:
The tests are mostly working, despite what it looks like here. There are some general issues however that we need to address to fully see their potential:

- The most generic problems arise from the `Code quality` check. This tells us that the code is not formatted correctly and that there are some issues like unsorted import blocks and unused variables. As mentioned before, tools like ruff (superseding isort and flake8) and black can be set up to automatically run whenever you save a file, eliminating most of these problems. 
- Some of the `Code quality` problems are not as easy to resolve. Things like unused variables need manual attention, even though they are simple: if you truly don't need the return value of a function, you can use an underscore: `_ = some_function()`. The mypy errors can be more tedious to resolve as we need to understand why the mentioned modules don't have the missing attributes and whether they should have or where else they should get them from.
- As mentioned before, the tests only contain a limited number of problems:
```python
ERROR message_ix_models/tests/model/water/test_cooling.py
ERROR message_ix_models/tests/model/water/test_utils.py
FAILED message_ix_models/tests/model/water/test_irrigation.py::test_add_irr_structure - AttributeError: 'dict' object has no attribute 'regions'
FAILED message_ix_models/tests/model/water/test_water_supply.py::test_map_basin_region_wat - AttributeError: 'dict' object has no attribute 'time'
FAILED message_ix_models/tests/model/water/test_water_supply.py::test_add_water_supply - ModuleNotFoundError: No module named 'your_module'
FAILED message_ix_models/tests/model/water/test_water_supply.py::test_add_e_flow - ModuleNotFoundError: No module named 'your_module'
```
- For the first two `FAILED`, the issue is the same: the actual function is using the context object incorrectly: 
```python
    info = context["water build info"]

    if "year" in context.time:
```
Are the [lines](https://github.com/iiasa/message-ix-models/blob/21daf50aef903d400c7cb0fea7cc4dd53bc5de44/message_ix_models/model/water/data/water_supply.py#L27-L29) from the actual code. I don't know if this should work, but it's treating `context` in two different ways: the first reminds me of a dictionary, the second of class. In your tests, you set up a mock `Context` object (which would normally be a class) as a dictionary, which is why the `info = ...` line works, but the second one complains: you cannot access dictionary data that way. Since I doubt you want to rewrite your actual code function to use both dict and `Context` for the `context` argument, I suggest to study how other [tests in this repo](https://github.com/iiasa/message-ix-models/blob/21daf50aef903d400c7cb0fea7cc4dd53bc5de44/message_ix_models/tests/model/test_macro.py#L27-L28) solve this problem. You should be able to use a `test_context` out of the box which you can then adapt.
- The last two `FAILED` result from trying to call a module called `your_module`, which doesn't exist. You probably want to replace `your_module` with `message_ix_model.model.water.data` in both cases.
- Finally, both `ERROR`s are due to missing imports. Both of these issues can probably be solved by addressing all mypy complaints, although one of them is trying to import message_data. This should not happen in message-ix-models and might be a sign that the clean-up process is still incomplete.

adrivinca:
@awais307 can you please update on what your commits did,  if anything is still missing and how you expect others to review? thanks

glatterf42:
@awais307, @adrivinca: the tests are almost working, you're on a good path there! Almost all errors seem to originate from the same problem: that you are not using a 'real' `Context` object. As pointed out above, using those is preferable since it makes the tests easier to understand and maintain and since you won't have to rewrite your actual code to deal with both `Context`s and `dict`s. 
With the latest commit, I adapted the test in `tests/model/water/test_cooling.py` to show you what I mean by that. We do have an object called `test_context` and we can use it just like that our test functions. Please take a look and try to adapt your other tests in a similar way, I suspect it will solve most of the remaining errors :)
If you don't want to write the same setup of the `text_context` all the time, you could introduce another fixture like it e.g. in the `tests/model/water/__init__.py`: you can define a function that takes the `test_context`, does whatever setup _all_ your tests frequently share, and return the modified `test_context`. You can then use the pytest decorator `@pytest.fixture(scope="function")` to make this function work identical to `test_context`. Then, you can use it in all your other tests as well :)

One test error might remain, though, and that is this one:
```python
___________________________________________ test_map_yv_ya_lt ___________________________________________

    def test_map_yv_ya_lt():
        periods = (2010, 2020, 2030, 2040)
        lt = 20
        ya = 2020
    
        expected = pd.DataFrame(
            {"year_vtg": [2010, 2020, 2020, 2030], "year_act": [2020, 2020, 2030, 2040]}
        )
    
        result = map_yv_ya_lt(periods, lt, ya)
    
>       pd.testing.assert_frame_equal(result, expected)
E       AssertionError: DataFrame are different
E       
E       DataFrame shape mismatch
E       [left]:  (8, 2)
E       [right]: (4, 2)

message_ix_models/tests/model/water/test_utils.py:124: AssertionError
```
The shape of the compared DataFrames won't be fixed by the different context, I assume. Please check that `expected` is really what we expect, e.g. `year_vtg` contains `2020` twice, is that on purpose?

One final note: I removed a few markers that told rust to ignore functions being too complex. Please take another look at those and try to reduce their complexity! In most cases, it is enough to take a part of the function that includes a lot of `if ... elif ... else ...` and move that to a new function that takes some input data and returns different output depending on the condition.

Please let me know if you need help with any of the above :)

glatterf42:
Hi @adrivinca and @awais307, I've taken another look at this PR. I started by addressing all code quality issues: applying some formatting and reducing code complexity. The latter worked quite well in parts because you already had all the puzzle pieces present: e.g. in `water_for_ppl.py`, your immensely complex `cool_tech()` function was in part only complex because you defined various helper functions _inside_ that same function. So by moving the definitions outside, the code quickly became less complex. 
Unfortunately, the same approach did not work as well for `infrastructure.py`. While moving stuff outside the main function also resolved the issue, there's still much room for improvements. Almost all steps in the main function that work with the input `DataFrame` seem to be very similar except for tiny variations, so according to the DRY principle (Don't Repeat Yourself), this code is most likely generalisable. 

Next, I took a look at the failing tests. For some of them, I was able to fix them myself, though even in those cases, there are some open questions. Please check e.g. the `TODO`s in `test_map_yv_ya_lt()`. 
For some cases, trying to use the `test_context` or otherwise fixing the tests led to problems to such a degree that I didn't know how to fix them. Please search for `FIXME` to find descriptions of these errors with probable causes and determine how you want to fix them.

Finally, your `reporting.py` contains these lines: https://github.com/iiasa/message-ix-models/blob/abcd6144752a742e9191cd4d51be3bd47de020e4/message_ix_models/model/water/reporting.py#L14-L36
`run_old_reporting()` is part of your main workflow (via `report_full()`), so `legacy_reporting` is actually required to be there. My version of mypy is already marking this as an error because `legacy_reporting` really cannot be `None`, so your code still requires message_data to be present, although no code in this repo should. Please find a solution for that, e.g. by not using the `legacy_reporting` at all and instead using the reporting system that message_ix_models provides. At the very least, you should add a clause/flag to `report_full` so that `run_old_reporting()` is only run if message_data is present. For this, you could reuse the `HAS_MESSAGE_DATA` flag that already exists in the code: https://github.com/iiasa/message-ix-models/blob/abcd6144752a742e9191cd4d51be3bd47de020e4/message_ix_models/util/common.py#L21

adrivinca:
Thanks a lot @glatterf42 for spending time on this and for the thorough overview.
I am going through the tests and the code, here the main changes I did

- Taking from another commit I did in the `leap-re-wks` branch, I replaced all the ` .append()` with `pd.concat([])`  and also another couple of parts where the code was obsolete
- concerning `test_map_yv` the test and functions work as intended. I just don't know why, when running locally the `reset_index()` in the main function does not seem to work when called from another module.
- I used `if HAS_MESSAGE_DATA:` in the `reporting.py` script to use the legacy reporting. Not sure the the reporting features in `message_ix_models` are outputting the same as legacy

Some FIXME are related to making a more general configuration for the module, and also maybe a testing config.
I would like to work on it for the SSP update, where I need anyway to change the code and for instance add the SSP config.
Therefore, I left some FIXME for the future, and I would merge them with the FIXME memo for now
I refer to the comments in `test_irrigation, test_water_supply`.

Other points are also open questions to me and I would like @awais307 to please intervene and fix.
in `test_water_supply`:
-  `test_add_e_flow` function does not make sense to me, how is the function supposed to work with fictitious regions if the files that get loaded are data files?

currently, the function is checking to match this df
```
   Region  value  year  time  Unnamed: 0  BCU_name
0  test_Region      1  2020  year           0  test_BCU
```
and df_sw in the function read_water_availability() before it breaks
```
     Region     years        value
0  test_BCU    Region  test_Region
1  test_BCU     value            1
2  test_BCU      year         2020
3  test_BCU      time         year
4  test_BCU  BCU_name     test_BCU
```
I think these functions need to be tested with actual existing data
- similar for `test_add_water_supply()`, please look at it, because it breaks, ut not sure if it is supposed to work with the input data that you give
- same for `test_map_basin_region_wat()` 
unfortunately I am not familiar with the `with   patch and mocking` approach @awais307  have you tried those tests you wrote?
When running pytest in the folder the al break
please also have a look to the annotation issue in `test_utils.test_add_commodity_and_level()`. and `test_utils.test_read_config`

glatterf42:
@awais3

glatterf42:
@awais307, I'm trying to keep the commit history linear by using `git rebase main`. A linear history means it will be very easy to understand how and why which changes were introduced and the changes will integrate with the rest of the code base neatly. So if you see that this branch is out-of-date with the `main` branch, I'd appreciate if you could also use `git rebase` instead of `git merge` :)
If you don't feel comfortable using `git rebase`, please check out e.g. [this tutorials](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase).

awais307:
> @awais307, I'm trying to keep the commit history linear by using `git rebase main`. A linear history means it will be very easy to understand how and why which changes were introduced and the changes will integrate with the rest of the code base neatly. So if you see that this branch is out-of-date with the `main` branch, I'd appreciate if you could also use `git rebase` instead of `git merge` :) If you don't feel comfortable using `git rebase`, please check out e.g. [this tutorial](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase).

okay, for some reason, my local tries to merge it whenever I fetch the changes.

glatterf42:
> > @awais307, I'm trying to keep the commit history linear by using `git rebase main`. A linear history means it will be very easy to understand how and why which changes were introduced and the changes will integrate with the rest of the code base neatly. So if you see that this branch is out-of-date with the `main` branch, I'd appreciate if you could also use `git rebase` instead of `git merge` :) If you don't feel comfortable using `git rebase`, please check out e.g. [this tutorial](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase).
> 
> okay, for some reason, my local tries to merge it whenever I fetch the changes.

Could you please try to configure your local git interaction not to do that, then? If you tell me how you interact with git locally, I might be able to help you. Do you use Github Desktop? GitKraken? The command line? 

To get the changes from up here to your system and overwrite the changes you have locally, you might need to perform something like a "force pull", which you can do following [these instructions](https://stackoverflow.com/a/8888015/23037943). Please let me know if you need any help with that :)

glatterf42:
Hi @awais307 :)
I took a closer look at the failing tests today and most of them should be fixed. Currently, I just see one local failure, which comes from the `test_build.py::test_build` which you recently added. Do you have any kind of confirmation that the water module functions should work with the `bare_res` scenario?

The test currently fails in the function `cool_tech`, which is called as part of `model/water/data/__init__.py::add_data()`, which in turn is part of `apply_spec()` as called in `model/water/build.py::main()`. The code complains that you can replace `df[[label1, label2]]` with something applied to `df[[label1, label2, label3]]`. If I adapt the lists of labels to fit, however, I run into another error a few lines later which states that we are trying to fill `None` values in `cooling_df` with values in `ref_input["technology"]`, but these values are all `None` themselves. So I suspect something is going wrong at a deeper level. 

I also suspect that adding a test for the function `cool_tech()` will shine a light onto why this other test is failing. 

In any case, I hope the examples for how to use the `test_context` that are now included in the test suite help you with figuring out how to use it for testing further cases :)

glatterf42:
I've added several new (superficial) tests that should increase the test coverage enough to make this PR mergeable :)
However, there are two main notes:

1. By superficial I mean that the tests don't check any actual resulting values, they just check that the results have roughly the correct form (e.g. column names are matching). This could be improved to actually make the tests reliable.
2. Some of the new tests are not passing. If marked them with `FIXME` and added notes about what I think is going wrong. Often, it's a question of "we want to use a Scenario to e.g. run the nexus reporting on, but for this we first have to create a Scenario properly, so that e.g. all cooling technologies can be found in the input parameter". I expect that you are much more familiar with how to do that, so I'd leave that up to you, @awais307 and @adrivinca.

awais307:
> I've added several new (superficial) tests that should increase the test coverage enough to make this PR mergeable :) However, there are two main notes:
> 
> 1. By superficial I mean that the tests don't check any actual resulting values, they just check that the results have roughly the correct form (e.g. column names are matching). This could be improved to actually make the tests reliable.
> 2. Some of the new tests are not passing. If marked them with `FIXME` and added notes about what I think is going wrong. Often, it's a question of "we want to use a Scenario to e.g. run the nexus reporting on, but for this we first have to create a Scenario properly, so that e.g. all cooling technologies can be found in the input parameter". I expect that you are much more familiar with how to do that, so I'd leave that up to you, @awais307 and @adrivinca.

Thanks a lot @glatterf42 . Very much appreciated for the contribution and support. Per the tests in demands, I think perhaps @adrivinca can confirm better since he did most of the recent adjustment with Zambia model. In principle, I would assume there is no need to read target rate function for a country case since those are meant for global basins.

adrivinca:
one of the problems that Fridolin mentioned has to do with [this function](https://github.com/iiasa/message-ix-models/blame/7da12e29bd5ecf1a564042bc38b763c38f2ba7af/message_ix_models/model/water/data/demands.py#L23)

```
def get_basin_sizes(basin: pd.DataFrame, node: str) -> Tuple[float, float]:
    """Returns the sizes of developing and developed basins for a given node"""
    temp = basin[basin["BCU_name"] == node]
    print(temp)
    sizes = temp.pivot_table(index=["STATUS"], aggfunc="size")
    print(sizes)
    return sizes["DEV"], sizes["IND"]
```
 I solved it by defining the sizes =0 in case they are empty before outputting.
Also the test function needed a different "SDG" in the case of the ZMB model, which I changed.

The other issue on `test_wateR_for_ppl::test_cool_tec` shows multiple issues that I adjusted:
- the test requires some parameters (input, historical_activity and historical_new_capacity ) to exist in the starting scenario, so I added them in the text
- the input RCP need to be "no_climate" instead of "no-climate" because of the input data files
- one part that Fridolin changed needed to be reverted
- one update in `cooling_fr()` related to pandas updates that I did in other branches was not included here. So I fixed it

adrivinca:
But i see some of the solutions that work for Windows do not work on Linux, for instance,
`message_ix_models/tests/model/water/data/test_water_for_ppl.py::test_cool_tec[no_climate] - KeyError: 'parent_tech'`
it should read a column from the incoming df, not sure why it gives an error. Any ideas @glatterf42 

Thanks!

glatterf42:
Hi @adrivinca, thanks for taking another look at this :)

I've tried addressing some of the issues here. As you noticed, there were some test failures that only occurred for tests that use ixmp and message_ix v3.6.0 (and older), so I've marked the function as needing a minimum version of 3.7.0. For 3.6.0 and older, our test suite forces pandas < 2.0.0, but one function in your tests uses a [parameter introduced with pandas 2.1.0](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.stack.html). This is totally fine, and if it happens again (or I forgot a function calling `add_sectoral_demand`), please apply the same changes that are now in https://github.com/iiasa/message-ix-models/pull/106/commits/feafa7edfd47d83eac78a15ba98263ef42686ef8.

Next, I noticed that you removed the updated assignment in `model/data/water_for_ppl.py:256` that I previously changed. Without my changes, I see a pandas error because 
```python
    ref_input[["value", "level"]] = ref_input[["technology", "value", "level"]].apply(
        missing_tech, axis=1
    )
```
tried to assign three colums of a dataframe to two columns (of _the same_ dataframe, which makes me think there should be a better way to simply update these two columns in the first place). This does not work because the number of columns on both sides of the assignment needs to match. So, assuming the dataframe you want to `apply` your `missing_tech` function to requires the `technology` column and you don't want to assign this column in `ref_input` again, I've now changed it to this instead:
```python
    ref_input[["value", "level"]] = ref_input[["technology", "value", "level"]].apply(
        missing_tech, axis=1
    )[["value", "level"]]
```
Selecting the columns we want to assign after they have been re-computed. For this to work, I had to also adapt `missing_tech` to return a pandas Series with correctly named columns.

With this, `test_build` still runs into an error: `NaN` values can't be converted to `int`. This seems to be the result of almost all dataframes in `cool_tech` being empty in this test, in particular `ref_input` in line 264. I'm not sure why this is the case and if this is supposed to work. So please clarify if `cool_tech` should be able to handle this case of empty dataframes, or if this is by mistake and these dataframes should never be empty. In the latter case, you could consider introducing a quick sanity check, either with `try...except...` to raise an error message or a warning if the dataframe is empty or with `assert ...` to ensure the dataframe is not empty (and raise an `AssertionError` otherwise).

Finally, we still have the error in `test_report_full`, which arises from the call to `legacy_report`. From my adaptation of the legacy reporting to message-ix-models, I remember that this error most likely comes from the fact that the default reporting is trying to calculate tables (the default tables, presumably), that you don't actually want to compute. Please take a look at your local reporting configuration and ensure this branch is computing the same tables. If this doesn't help, @OFR-IIASA is the expert on message_data-style reporting and can probably quickly point you to what's wrong if you ask nicely :pray:

glatterf42:
For completeness' sake: there is an error about a `dimensionless` unit. This is completely unrelated to this PR and will be fixed upstream in pint at some point. Please disregard for this PR :)

adrivinca:
Thanks @glatterf42 , sorry for the long delay, but it's hard to find time for this. 
Concerning the code you changed, I understood your changes from before. but on my machine it behaved differently, like as the function needed 2 inputs and had 3 outputs... maybe some differences in packages?

concerning the reporting, I did not change it to the legacy version, which I never used, so I was expecting that whoever updates the code also make sure it works...
Still, I will dedicate time to solving those two issues over the next two days. But if some errors remain too complicated I would ask to please proceed with merging asap.
There are updates to be done that might drastically change the module, and the tests accordingly. So I would rather keep some tests commented and get back to them after the updates.
Cheers

glatterf42:
No worries, thanks for dedicating some time now :)

For the packages, the error might indeed come from different versions. Trying to work with this PR from a new virtual environment (as you already indicated you would on Slack) sounds like the best way to solve the issue :)

Please help me understand what is going on with the reporting, then: 
https://github.com/iiasa/message-ix-models/blob/cf687da50760239cf72d488a662fbc08bd583907/message_ix_models/model/water/reporting.py#L20C1-L32C6
This function is called `run_old_reporting()` and, looking at the blame, was called so and written almost entirely by you. It mainly calls the reporting system from message_data, which is called "legacy_reporting" here.
Back in the original migration PR, you stated that your code relies on the `report()` function from message_data: https://github.com/iiasa/message-ix-models/pull/88#issuecomment-1436986943
@khaeru then suggested a workaround to limit importing this function to only those systems that have message_data installed. 
This is not really a good solution, though, since our open-source modules should not rely on closed-source modules. Thus, #159 took the version of `report()` from message_data that was on `ssp_dev` at the time and plugged it into message-ix-models. We don't make any guarantees that this works with any code except for one: this reporting function works with the global baseline snapshot we published on Zenodo. 
Still, with https://github.com/iiasa/message-ix-models/pull/106/commits/1ecad356ebb9eca47b9e062c36b7ad96d136b337, I migrated the water-report-`legacy_report()` call to the new message-ix-models-`legacy_report()`. If this doesn't work out of the box, you essentially have a few options:

- Migrate back to the conditional import from message_data
- Adapt the `legacy_report` functionality here so that it satisfies both our working test and your new one
- Migrate your code so that it only uses the new reporting feature

Arguably, the first option is the least work and the worst style, while the last option is the most work, best style, and highly appreciated by everyone else who would also benefit from this.
Please let me know if I got that right and what you intend to do :)

Finally, please note that I force-pushed this branch to integrate the latest changes from `main`. So please `git fetch` this branch if you have local changes and `git rebase` your local changes on top of this version here. My code improvement tools signal that some things are not working correctly immediately, so I will also draft a small commit or two addressing these issues.

khaeru:
Looking at the test failures here:

----

```
ERROR message_ix_models/tests/tools/test_wb.py::test_get_income_group_codelist - requests.exceptions.HTTPError: 502 Server Error: Bad Gateway for url: http://api.worldbank.org/v2/sdmx/rest/codelist/WB/
ERROR message_ix_models/tests/tools/test_wb.py::test_assign_income_groups[R12-population-None] - requests.exceptions.HTTPError: 502 Server Error: Bad Gateway for url: http://api.worldbank.org/v2/sdmx/rest/codelist/WB/
ERROR message_ix_models/tests/tools/test_wb.py::test_assign_income_groups[R12-count-None] - requests.exceptions.HTTPError: 502 Server Error: Bad Gateway for url: http://api.worldbank.org/v2/sdmx/rest/codelist/WB/
ERROR message_ix_models/tests/tools/test_wb.py::test_assign_income_groups[R12-population-0] - requests.exceptions.HTTPError: 502 Server Error: Bad Gateway for url: http://api.worldbank.org/v2/sdmx/rest/codelist/WB/
FAILED message_ix_models/tests/tools/test_wb.py::test_make_map - requests.exceptions.HTTPError: 502 Server Error: Bad Gateway for url: http://api.worldbank.org/v2/sdmx/rest/codelist/WB/
```
See the commit on #213 with the subject "Use HTTPS to access World Bank SDMX endpoints / Work around khaeru/sdmx#191". You could cherry-pick this commit onto the current branch.

----

```
FAILED message_ix_models/tests/model/transport/test_report.py::test_plot_simulated[stock-ldv] - genno.core.exceptions.MissingKeyError: required keys ('fuel economy:nl-m:ldv+ref',) not defined
FAILED message_ix_models/tests/model/transport/test_report.py::test_iamc_simulated - genno.core.exceptions.MissingKeyError: required keys ('fuel economy:nl-m:ldv+ref',) not defined
```
These I am also seeing in #213 and struggling to reproduce locally, so I do not yet have a fix to merge or for you to cherry-pick. You could mark these with:

```python
@pytest.mark.xfail(reason="Temporary, for #106")
```
…and make a note in the PR description that you have done this.

----

Let me look briefly at the other two…

khaeru:
```
FAILED message_ix_models/tests/model/water/test_build.py::test_build - pandas.errors.IntCastingNaNError: Cannot convert non-finite values (NA or inf) to integer: Error while type casting for column 'year_act'
```

Looking at the full pytest output for this test, we see:
```python
        # Code runs on the bare RES
>       build(context=test_context, scenario=scenario)

message_ix_models/tests/model/water/test_build.py:31: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
message_ix_models/model/water/build.py:268: in main
    build.apply_spec(scenario, spec1, partial(add_data, context=context), **options)
message_ix_models/model/build.py:155: in apply_spec
    result = data(scenario, dry_run=dry_run)
message_ix_models/model/water/data/__init__.py:64: in add_data
    add_par_data(scenario, func(context), dry_run=dry_run)
message_ix_models/model/water/data/water_for_ppl.py:264: in cool_tech
    .combine_first(ref_input.set_index("technology"))

... [internal to pandas] ...

/opt/hostedtoolcache/Python/3.12.5/x64/lib/python3.12/site-packages/pandas/core/dtypes/astype.py:101: in _astype_nansafe
    return _astype_float_to_int_nansafe(arr, dtype, copy)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

values = array([nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan,...nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan])
dtype = dtype('int64'), copy = True

    def _astype_float_to_int_nansafe(
        values: np.ndarray, dtype: np.dtype, copy: bool
    ) -> np.ndarray:

...

E           pandas.errors.IntCastingNaNError: Cannot convert non-finite values (NA or inf) to integer: Error while type casting for column 'year_act'
```

This indicates that all-NaN values are present in a "year_act" column in one of the variables used in this statement: https://github.com/iiasa/message-ix-models/blob/73ab3e4e3343ce7d6b425ed0ce4a9402c9eb1012/message_ix_models/model/water/data/water_for_ppl.py#L261-L266

You could debug by displaying these variables just before this statement and identifying why this happens.

khaeru:
```
FAILED message_ix_models/tests/model/water/test_report.py::test_report_full - SystemExit: 1
```
As far as I can see: https://github.com/iiasa/message-ix-models/blob/73ab3e4e3343ce7d6b425ed0ce4a9402c9eb1012/message_ix_models/tests/model/water/test_report.py#L12-L46

…this test sets up a new, completely empty scenario; builds MESSAGEix-Nexus on top of that; solves it; and then runs the `report_full()` function, which in turn invokes the legacy reporting.

Looking at the exception raised [here](https://github.com/iiasa/message-ix-models/actions/runs/10503430449/job/29096893089?pr=106#step:9:3797) in `.report.legacy.postprocess._retr_land_act()`, we can conclude a few things:
- The legacy reporting code expects some values in the LAND variable, and cannot (or just refuses to) proceed without them.
- This is not documented.
- The code calls `sys.exit(1)` directly instead of the Pythonic behaviour of raising an exception.

There are various ways (not mutually exclusive) to approach this:
1. Expand the test such that `LAND` is non-empty in the solution. This could involve:
   a. Create minimal ("dummy") parameter values that achieve this.
   b. Build on the snapshot scenario instead.
2. In `.water.report`, adjust:
   https://github.com/iiasa/message-ix-models/blob/73ab3e4e3343ce7d6b425ed0ce4a9402c9eb1012/message_ix_models/model/water/report.py#L26-L31
   to:
   ```python
   try:
      legacy_report(...)
   except SystemExit:
      log.warning("Legacy reporting failed")
   ```
   (However please note that `message_ix_models.report.report(...)` is the preferred way to invoke the legacy reporting. That could also be the proper place for this protective code. If you decide to use this approach, please help me in and I can jump in.)
3. Improve the documentation of the conditions where `.report.legacy` is usable.

adrivinca:
Thanks Paul and Frido. 
I am looking into the build issue.

Concerning reporting, I think the legacy reporting does not work if a scenario is empty, The scenario requires the globiom matrix emulator (that's why we have LAND related error).
So if you want to make sure it is somehow testable, maybe who migrated the legacy reporting could do a test that tests a normal ENGAGE (or SSP) scenario first in another PR. Then that could be taken as an example for other branches. Otherwise I am afraid we are charging too much stuff on this PR

I also tried to only run the water reporting and I think it also work only with actual scenarios. Some operation will not work if there is not output (I already get an error from "Reporter")
So, if possible I would not test it right now - eventually just test the imports of the reporting file- and leave it for the next update to prepare a nexus_bare small scenario for testing. But I don't want to do it now because there are major changes coming.

Does it make sense?

glatterf42:
Thanks for your analysis @khaeru, I have cherry-picked your commit as suggested. 
I have also tried to address the numerous mypy issues, of which I don't understand how they were not caught with our code quality check here before. Sometimes, these approaches include 'best guesses' as to what the code should do and I tried to mimic that, but I might have been wrong, so @adrivinca please check that the code still works as intended.

I was the one who migrated the legacy reporting. As far as I know, the snapshot on Zenodo is a baseline snapshot for ENGAGE. Other than that, I don't know what a "normal scenario" is supposed to be and where I could get one; I certainly don't have the time to build one from scratch.
Agreed, making drastic changes to `legacy_report()` is out of scope for this PR. However, if I remember correctly, `legacy_report()` tried to run different "reporting tables" including those for LAND output. These could be disabled pretty easily, but I don't remember how exactly we did that. It might be documented in https://github.com/iiasa/message-ix-models/pull/159, though.

In general, as @khaeru suggested, you are free to mark certain tests as `XFail` or so, but please record in a new issue which tests you marked and why and try to address them soon (if after your upcoming major changes).

khaeru:
> So, if possible I would not test it right now - eventually just test the imports of the reporting file- and leave it for the next update to prepare a nexus_bare small scenario for testing. But I don't want to do it now because there are major changes coming.
> 
> Does it make sense?

Yes, this makes sense and is a fair strategy. As @glatterf42 says, you can still retain the `test_report_full()` that is added on this branch, as a starting-point for that future work. Just mark it XFAIL so that pytest understands that is expected:

```python
@pytest.mark.xfail(reason="Incomplete")
def test_report_full(...):
    ...
```

glatterf42:
We should now be at the same two errors we had before the mypy commit, I'll leave it at that for now.

adrivinca:
Apologies, I was one week on leave and I left it pending. Thanks again for your effort.
I was looking at the `test_build.py` issue and simply the script cannot work because the scenario given as input is basically empty, for instance, it does not have an input parameter. the script runs over all the other scrips (like water_for_ppl, demands.py and so on) and, like the test that were testing those functions, it would need some parameters to start with.

for instance in water_for_ppl line 242, `ref_input `is empty n the test scenario
```
    ref_input: pd.DataFrame = scen.par(
        "input", {"technology": cooling_df["parent_tech"]}
    )
```
as a consequence `input_cool` is empty and the code gives error, line 262
```
input_cool = (
        cooling_df.set_index("parent_tech")
        .combine_first(ref_input.set_index("technology"))
        .reset_index()
    )
```

I could try to add manually some dummy parameters as did in [test_water_for_ppl.py](https://github.com/iiasa/message-ix-models/blob/nexus-upd/message_ix_models/tests/model/water/data/test_water_for_ppl.py) and see if it works. But for this I have a question. I see the starting scenario in `test_build` is defines as `scenario = testing.bare_res(request, test_context)`, while in `test_demands` and `test_water_for_ppl` it is defined as `s = Scenario(**scenario_info)`. What is the difference and what is recommendable?

Anyway I am not sure the build would work with a semi-empty scenario. sometimes, specific data is needed. Also, I would assume that if the functions that build calls are already tested I would check something

adrivinca:
Hi, as motivated in the previous messages, I marked two tests with @pytest.mark.xfail(reason="Temporary, for #106"), namely
test_report_full in test_report.py
and test_build in test_build.py
as they both require a fully parametrized scenario as a starting point, which is not the case for the "bare" scenario for tests.

Since we are about to do major revision and update to the module, we decided to postpone investing further time in those tests.

all tests seem to pass for py3.12, not for py3.11, is that a problem?
otherwise I would ask to kindly merge.
thank you very much for the great support!

glatterf42:
All tests seem to be passing now :)
So in light of the above discussion and with the explicit promise to write tests for the water module as soon as possible after your upcoming major revision, I think we can merge this :)

PR #107, opened 2023-07-04
Include message_doc

### LINKED ISSUE #105: Integrate message_doc
As discussed in the weekly meeting today, we want to integrate the documentation of the GLOBIOM model that is currently living under message_doc ([docs](https://docs.messageix.org/projects/global/en/latest/), [rtd](https://readthedocs.com/projects/iiasa-energy-program-message-doc/), [github](https://github.com/iiasa/message_doc)) in the documentation of message-ix-models. But where/how exactly should that happen? 
We should 
- [ ] leave behind redirects for people who saved bookmarks to the current docs
- [ ] do this before July 24 (ideally, first brown out), but definitely before October (full deprecation from RTD side of building docs w/o v2 config files)

glatterf42:
Closes #105. This PR moves all files from [message_doc](https://github.com/iiasa/message_doc/tree/main) to `doc/global/` or their appropriate places (e.g., merging the `doc/_static` directory).

### Steps run so far
This follows the steps outlined in #86. In a fresh clone of message_doc, I ran
```bash
git filter-repo --path-rename :doc/global/
git filter-repo --path-rename doc/global/_static/:doc/_static/
git filter-repo --path doc/_static/combined-logo-white.png --invert-paths
```
In a fresh clone of message-ix-models, this was continued via
```bash
git remote add message_doc ../message_doc
git remote update
git checkout -b tmp/message-doc
git merge message_doc/main --allow-unrelated-histories
git checkout main
git checkout -b include/message-doc
git merge tmp/message-doc --allow-unrelated-histories
git push -u origin include/message-doc
```

### Further steps mentioned by @khaeru and me:

- In #110:
  - Merge necessary parts of the sphinx config to the existing one, delete the one in `globiom/`
  - Integrate index page in the same way.
  - Integrate bibliography in the same way.
  - Ensure the text of the docs is still coherent given its new location.
- [ ] Leave behind redirects from old location of the docs (RTD-related)

## How to review

- Build the documentation and skim through the GLOBIOM section to ensure it still makes sense.

## PR checklist

- ~[ ] Add or expand tests; coverage checks both ✅~ No functionality added
- [x] Update doc/whatsnew.


khaeru:
Excellent, thanks for getting this going.

Some thoughts:
- "MESSAGEix-GLOBIOM" is the name used for the global model family developed by ECE. However, "GLOBIOM", on its own, is a separate model maintained by IIASA BNR. "MESSAGEix-GLOBIOM" only contains an emulator (of sorts) for results from the actual GLOBIOM model.

  I am not sure whether using "globiom" as the sub-path invites confusion. At the moment, the message_doc build appears at:

  https://docs.messageix.org/global/
  https://docs.messageix.org/projects/global/

  I would suggest to use the same fragment, so that this appears at:

  https://docs.messageix.org/models/global/
  https://docs.messageix.org/projects/models/global/

  However, perhaps best to raise to the whole team in Slack or at the MESSAGE meeting, and ask what is easiest to communicate and least confusing.

- I activated the RTD build for this branch. For the moment the root page of the added files appears at: https://docs.messageix.org/projects/models/en/include-message-doc/globiom/index.html —however this would change per the above comment and the TODOs to integrate the ToCs.
- I realized we use a logo variant with a purple centre for the message_doc site (vs. yellow for the current repo). With this integration, the purple one will become unused, and could be reused elsewhere later.

glatterf42:
Force push and edit in the description to accommodate the new location of the docs in `doc/global/` rather than `doc/globiom/`.

khaeru:
Something to ponder here: the commit messages in the full history don't conform to style we prefer for message-ix-models, and some are totally uninformative. Should we squash these, rewrite them (there are a lot), or something else?

glatterf42:
Assuming that we'd have to go through the 1538 commits manually, that seems like it requires lots of time, even if dealing with each commit only takes a few minutes (say 4, then that still results in 100 hours to spend on this). That is also assuming that the counted line changes are correct, which seems odd as there are no deletions. Because if each commit contains many more changes compared to the roughly four lines added on average with the current numbers, the 4-minute-assumption becomes much less likely.
So while I agree that it would be nice to have a clean history here, I'm not sure it's worth the effort. The question, to me, seems to be: how often do people actually go looking to the git history in search of some specific change they need to pin down?
And even if that's more than a couple times per year, they should be able to find the commit introducing the changes they are interested in, so another question is: between knowing the author and being able to look at preceding and following commits, how much more value would a clean commit message provide?

khaeru:
> So while I agree that it would be nice to have a clean history here, I'm not sure it's worth the effort. The question, to me, seems to be: how often do people actually go looking to the git history in search of some specific change they need to pin down?
> And even if that's more than a couple times per year, they should be able to find the commit introducing the changes they are interested in, so another question is: between knowing the author and being able to look at preceding and following commits, how much more value would a clean commit message provide?

Good questions!

I think I'm mainly concerned that the history of the message-ix-models code itself is not obscured or made less useful by the 1538 additional commits.

We should also consider:
- What will be the fate of the iiasa/message_doc repo? I think we will likely eventually put it into "archive" state. If so, we could put a note somewhere to consult the history/blame (e.g. https://github.com/iiasa/message_doc/blame/main/index.rst) of the archived repo, which will remain available even if we chose to "squash and merge" here.
- The releases or tags in message_doc are used with RTD features to provide access to older versions of the MESSAGEix-GLOBIOM documentation. (I don't know if/how often these are consulted, if ever.) Do we want to do this in the same way? Or could we e.g. keep and link the PDF builds of those older versions, on the off-chance that people want to read those?

Still not sure what the choice will be, we can consider as work proceeds on #110.

khaeru:
I've just force-pushed this branch and rebased the one for #110 on it.

Some details:
- I prepared this new branch using the code in #174. I used [replacements.txt](https://github.com/iiasa/message-ix-models/files/15017749/replacements.txt) as one input. This generated [rebase-todo.in](https://github.com/iiasa/message-ix-models/files/15017763/rebase-todo-in.txt), from which I prepared [rebase-todo.txt](https://github.com/iiasa/message-ix-models/files/15017756/rebase-todo.txt) ([rebase-todo.patch](https://github.com/iiasa/message-ix-models/files/15017769/rebase-todo-patch.txt)).
- We now see at the [network graph](https://github.com/iiasa/message-ix-models/network) that the "Initial commit" of message-ix-models *per se* is the earliest in the history, as expected. Instead of a separate initial commit for the integrated message_doc history, the other repo's initial commit appears as the first commit on this branch, as expected.
- Taking advantage of the new features in the migration tool:
  - Commit messages are automatically filtered to start with an upper-case and to change verb tense to match the style preferred for this repo: for instance, "added new file" becomes "Add new file".
  - Consecutive commits with (a) the same commit message (things like "Update 8_special_features_of_the_matrix_generator.rst") and (b) the same author name are squashed together. This reduces the history by about a third, from 1537 to 1040 commits.

glatterf42 (commented):
Thanks, looks good to me. In particular the linear history :)
Since I'm technically the author of this PR, I can't formally approve it.

khaeru (approved):

khaeru:
The 5 failures of "pytest" jobs are all the same consequences of the ixmp4 bug we are tracking in #173, so not a barrier to merging this.

PR #124, opened 2023-09-13
Delete missing/multiple/duplicate caches in "pre-commit" job

khaeru:
"gh cache list" may occasionally give zero results; multiple results; or multiple results with the same cache key. Allow all of these to succeed with exit code 0.

## How to review

Not required.

## PR checklist

- [x] Continuous integration checks all ✅
- [x] ~Add or expand tests;~ coverage checks both ✅
- ~Add, expand, or update documentation.~ N/A; CI changes only.
- [x] Update doc/whatsnew.

PR #127, opened 2023-09-18
Fix bash syntax in "Code quality" CI job

khaeru:
The change introduced in #124 contained a bash syntax error; this fixes.

## How to review

## PR checklist

- ~Add, expand, or update documentation.~ N/A; CI changes only
- ~Update doc/whatsnew.~

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -113,7 +113,7 @@ jobs:
 
     - name: Force recreation of pre-commit virtual environment for mypy
       if: github.event_name == 'schedule'  # Comment this line to run on a PR
-      run: gh cache list -L 999 | cut -f2 | grep pre-commit | xargs -I{} 'gh cache delete "{}" || true'
+      run: gh cache list -L 999 | cut -f2 | grep pre-commit | xargs -I{} gh cache delete "{}" || true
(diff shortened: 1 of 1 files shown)


khaeru:
[Here](https://github.com/iiasa/message-ix-models/actions/runs/6222001272/job/16885059036#step:4:7) is output from a run triggered by a temporary commit on this branch (now removed). The output:
```
X Could not find a cache matching pre-commit-3||3dc301320e6dab9741dd1dfa05e0841c73bf77bf413389c45f8637dae3c8a56c in iiasa/message-ix-models
```
…occurs because the bash code `gh cache list -L 999 | cut -f2 | grep pre-commit` generates:
```
pre-commit-3||3dc301320e6dab9741dd1dfa05e0841c73bf77bf413389c45f8637dae3c8a56c
pre-commit-3||3dc301320e6dab9741dd1dfa05e0841c73bf77bf413389c45f8637dae3c8a56c
```
…that is, duplicate lines for exactly the same cache ID. (I have no idea why this happens; maybe some artifact of how GitHub's caching infrastructure works.)

The bash code `xargs …` then tries to delete this, twice: the first time succeeds, and the second fails because the cache item has already been deleted. So ` || true` ensures that the job step overall succeeds.

PR #146, opened 2024-01-19
Remove Python 3.10-exclusive kw_only dataclass argument temporarily

glatterf42:
As discovered by @behnam-zakeri, our current code should either require python >= 3.10, or remove the `dataclass()` argument `kw_only` until this is the case.
Should we leave a note somewhere that we strongly encourage users to still use the syntax you were trying to enforce, @khaeru?

Please note: this PR leaves a TODO comment to come back to it. The other changes are small notes that mypy complained about. 

## How to review

## PR checklist

- ~[ ] Add, expand, or update documentation.~
- [x] Update doc/whatsnew.

### DIFF
--- a/message_ix_models/util/scenarioinfo.py
+++ b/message_ix_models/util/scenarioinfo.py
@@ -61,8 +61,9 @@ class ScenarioInfo:
     .Spec
     """
 
+    # TODO: give this field kw_only=False once python 3.10 is the minimum version
     # Parameters for initialization only
-    scenario_obj: InitVar[Optional["Scenario"]] = field(default=None, kw_only=False)
+    scenario_obj: InitVar[Optional["Scenario"]] = field(default=None)
     empty: InitVar[bool] = False
 
     platform_name: Optional[str] = None
--- a/message_ix_models/util/scenarioinfo.py
+++ b/message_ix_models/util/scenarioinfo.py
@@ -89,7 +89,7 @@ class ScenarioInfo:
     #: :obj:`True` if a MESSAGE-MACRO scenario.
     is_message_macro: bool = False
 
-    _yv_ya: pd.DataFrame | None = None
+    _yv_ya: Optional[pd.DataFrame] = None
 
     def __post_init__(self, scenario_obj: Optional["Scenario"], empty: bool):
         if not scenario_obj:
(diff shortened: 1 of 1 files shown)


glatterf42:
Edit: the `ScenarioInfo.scenario_obj` field had the `kw_only` argument, too, and I left it in at first because I thought it wouldn't hurt to keep it for the future (since it was set to its default `False` value. But python < 3.10 might still complain about not knowing this argument, so I commented it out as well.

behnam-zakeri:
> Edit: the `ScenarioInfo.scenario_obj` field had the `kw_only` argument, too, and I left it in at first because I thought it wouldn't hurt to keep it for the future (since it was set to its default `False` value. But python < 3.10 might still complain about not knowing this argument, so I commented it out as well.

Thanks @glatterf42. Indeed, that instance needs to be commented/removed too.

glatterf42:
Thanks for confirming. I think the codecov failure is due to the two pushes to this branch being in quick succession, this sometimes leads to coverage reports being not properly collected. The apparently uncovered lines are in a file I didn't touch, at least.

message_ix_models/util/scenarioinfo.py
```diff
@@ -87,7 +89,7 @@ class ScenarioInfo:
     #: :obj:`True` if a MESSAGE-MACRO scenario.
     is_message_macro: bool = False
 
-    _yv_ya: pd.DataFrame = None
+    _yv_ya: pd.DataFrame | None = None
```
khaeru, line 92: Note this is only valid from Python 3.10. To keep compatibility with 3.9 and 3.8, we need to use `Optional[pd.DataFrame]`.

khaeru (approved):
This seems fine; however please see the one comment about `Optional`.

>  Should we leave a note somewhere that we strongly encourage users to still use the syntax you were trying to enforce?

We could in fact substantially expand the docs of ScenarioInfo with narrative explanation and examples of the many ways the class can be/is used in message-ix-models and message_data. For instance, see how doc/api/util.rst has extra content on how to use the Context class.

Whether to do that in this PR or another, I leave to you.

glatterf42:
Thanks, that would have defeated the purpose of this PR. 

I'd leave the expansion of the docs for another PR to avoid scope creep.

Issue #147, opened 2024-01-22
Too similar file names on macos


glatterf42:
As pointed out by @measrainsey last year and now by @ravitby, this repository currently contains both [demand.csv.gz](https://github.com/iiasa/message-ix-models/blob/af092b1cfc7824019b5991319b5c557a096826a1/message_ix_models/data/test/MESSAGEix-GLOBIOM_1.1_R11_no-policy_baseline/demand.csv.gz) and [DEMAND.csv.gz](https://github.com/iiasa/message-ix-models/blob/af092b1cfc7824019b5991319b5c557a096826a1/message_ix_models/data/test/MESSAGEix-GLOBIOM_1.1_R11_no-policy_baseline/DEMAND.csv.gz) in `data/test/MESSAGEix-GLOBIOM_1.1_R11_no-policy_baseline`. This seems to cause git conflicts on macos systems as they don't distinguish between capitalized and non-capitalized file names.

I'm not immediately sure how we can remedy this, possibly by renaming a file, but I'm not sure if the exact names are somehow important (e.g., correspond to some model parameter and need to be exactly the way they are for the file to be found).

ravitby:
The result of the duplicate files is that there is an always a change in the repo and therefor it's impossible to pull new changes.
This is the output message for cloning the repo:

```Cloning into 'message-ix-models'...
remote: Enumerating objects: 25083, done.
remote: Counting objects: 100% (3728/3728), done.
remote: Compressing objects: 100% (1448/1448), done.
remote: Total 25083 (delta 2505), reused 2374 (delta 2278), pack-reused 21355
Receiving objects: 100% (25083/25083), 19.41 MiB | 11.41 MiB/s, done.
Resolving deltas: 100% (15969/15969), done.
Updating files: 100% (565/565), done.
Filtering content: 100% (353/353), 89.53 MiB | 5.25 MiB/s, done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'message_ix_models/data/test/MESSAGEix-GLOBIOM_1.1_R11_no-policy_baseline/DEMAND.csv.gz'
  'message_ix_models/data/test/MESSAGEix-GLOBIOM_1.1_R11_no-policy_baseline/demand.csv.gz'
```

glatterf42:
[...]

PR #149, opened 2024-01-22
Delete unused DEMAND data file

glatterf42:
As suggested by @khaeru in #147, this PR removes `message_ix_models/data/test/MESSAGEix-GLOBIOM_1.1_R11_no-policy_baseline/DEMAND.csv.gz` since it's not mentioned anywhere in the current code base.

Trying to run the tests locally, I got stuck on `test_snapshot`, which might be most crucial here. `test_cli` also complained that `ixmp.utils` has no attribute `parse_url`, so I updated this to `ixmp.util`.

## How to review

- Do you think we need to leave a note for when we are adding code/data again that need to reference both `demand` and `DEMAND`?

## PR checklist

- ~[ ] Add or expand tests; coverage checks both ✅~ All tests still working
- ~[ ] Add, expand, or update documentation.~ Do we mention this file somewhere? A [quick search](https://docs.messageix.org/projects/models/en/latest/search.html?q=DEMAND&check_keywords=yes&area=default) yields nothing.
- ~[ ] Update doc/whatsnew.~ This file was not used, so no need.

### DIFF
--- a/message_ix_models/data/test/MESSAGEix-GLOBIOM_1.1_R11_no-policy_baseline/DEMAND.csv.gz
+++ b/message_ix_models/data/test/MESSAGEix-GLOBIOM_1.1_R11_no-policy_baseline/DEMAND.csv.gz
@@ -1,3 +1,0 @@
-version https://git-lfs.github.com/spec/v1
-oid sha256:3334223695badec19f0a01d9063e77cbb88fcad70090b97be3a34b90e239790d
-size 7351
--- a/message_ix_models/util/context.py
+++ b/message_ix_models/util/context.py
@@ -371,7 +371,7 @@ class Context(dict):
                 )
 
             self.core.url = url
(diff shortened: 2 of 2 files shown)


khaeru (approved):
Thanks, LGTM so people can rebase more easily.

> Do we mention this file somewhere? A [quick search](https://docs.messageix.org/projects/models/en/latest/search.html?q=DEMAND&check_keywords=yes&area=default) yields nothing.

This file is produced by the [`unpack()`](https://docs.messageix.org/projects/models/en/latest/api/model-snapshot.html#message_ix_models.model.snapshot.unpack) function, but the testing and reproducibility docs don't cover non-package data that is stored in the repo and used for testing.

> Do you think we need to leave a note for when we are adding code/data again that need to reference both `demand` and `DEMAND`?

I think we will notice the need to address this when tests of such added code fail, which should lead us back to this issue/PR.

PR #150, opened 2024-01-25
Save minutes by avoiding duplicate test runs

glatterf42:
This PR removes the trigger for CI runs on pushed to main to avoid duplicate test runs. This avoids duplicate runs.

In order for this to work properly, we should make sure that

- [x] main can only be reached via PRs
- [x] PRs need to be up to date with main
- [x] PRs require several tests to be successful before merging

We might also want to 

- [ ] expand the number of tests we require to be successful

All of which can be done via the [settings -> branches](https://github.com/iiasa/message-ix-models/settings/branch_protection_rules/24406637).

## How to review

## PR checklist

- ~[ ] Add or expand tests; coverage checks both ✅~ Just changing how often tests are run.
- ~[ ] Add, expand, or update documentation.~ Just changing how often tests are run.
- ~[ ] Update release notes.~ Just changing how often tests are run.


khaeru:
> We should include checks for message_ix v3.8.0 (or only use main)

Yes, including both the latest release *and* main should be our general rule.

> and could try using py3.12, too.

Worth a try! I not yet sure if all of the direct dependencies here are compatible with Python 3.12, but we could find out.

khaeru (approved):

khaeru:
> We might have to re-trigger a check evaluation to get rid of the '3.11-required' checks and update the branch protection rule

I find if I modify the required checks, save settings, and then refresh the PR page, then the new requirements are shown without the need to re-run any jobs.

However it looks like there are several failures, so one option would be to park some notes in an issue to be addressed separately and limit this PR to adding ixmp/message_ix 3.8.0. Or try to also address them in this PR—up to you.

glatterf42:
This PR now refines the pytest workflow usage of python-version. A lot of the failures originated from the incompatibility of (ixmp, message_ix) < 3.8.0 with python 3.12 and we are not requiring this compatibility anywhere, so this workflow shouldn't require it either.

khaeru:
- Pint does include py.typed, for about 3 years: https://github.com/hgrecco/pint/commit/c60e1af833391a60a82f3c47460fca899e220de5
- The approach for genno is to include it in the list of packages in pre-commit: https://github.com/khaeru/genno/blob/main/.pre-commit-config.yaml#L10 —there is no exclusion in pyproject.toml.
- In the .pre-commit-config.yaml additional dependencies for mypy, there are two options:
  - If a package like `lxml-stubs` or `types-requests` exists, then use this. This does not install e.g. `lxml` or `requests` or any of their dependencies.
  - Otherwise, use packages directly, like `xarray`. This results in the full, actual package being installed in the pre-commit venv used for mypy—thus also all of its dependencies.
- The ixmp .pre-commit-config.yaml includes genno, which has a mandatory dependency on pint, so pint is installed in the pre-commit venv used for mypy.
- Here, the .pre-commit-config.yaml includes ixmp, but there is no direct dependency on pint, so pint is not installed. Two options would be:
  - Add pint directly to .pre-commit-config.yaml (preferred)
  - Change the ixmp line to something like `ixmp[report] @ git+https://github.com/iiasa/ixmp.git@main`. IMHO indirect dependencies are best for pyproject.toml/PyPI, but direct ones are fine for mypy.

glatterf42:
Thanks for the explanation. If I didn't misunderstand it entirely, f7d54e3 should be exactly the change you suggested. However, this doesn't solve the problem here (and probably won't for message_data, either, since the error message (also for message_data) reads:
```bash
message_data/testing.py:7: error: Skipping analyzing "pint": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```

glatterf42:
Okay, so for the last CI check before the force-push, we had: https://github.com/iiasa/message-ix-models/blob/cba07b2e7864ad16bb20367ff7ece11a4629776c/.github/workflows/pytest.yaml#L111 commented out and https://github.com/iiasa/message-ix-models/blob/cba07b2e7864ad16bb20367ff7ece11a4629776c/.pre-commit-config.yaml#L11-L12 appended with `; python -m pip list` at the end of the quotation marks. For some reason, this made the [code quality check successful](https://github.com/iiasa/message-ix-models/actions/runs/7798071731/job/21266015780), but it's failing again now.

glatterf42:
For some reason, the Codecov action v4 keeps failing in this repository, while it works fine with ixmp and message_ix and I don't remember setting anything up differently between them.

glatterf42:
> For some reason, the Codecov action v4 keeps failing in this repository, while it works fine with ixmp and message_ix and I don't remember setting anything up differently between them.

At least this seems clear now: I forgot to pass the CODECOV_TOKEN secret to the action. On the other repos, the PR is coming from a fork, which triggers a tokenless upload.

glatterf42:
There it is again: adding `; python -m pip list` to the pre-commit-config bash mypy call makes the code quality CI check successful. Let's see if the same happens for message_data.

glatterf42:
@khaeru Gotta be honest here: I don't understand what's going on here, but the tests are successful here and over at https://github.com/iiasa/message_data/pull/533, so it seems like we can adapt the required-successful checks to check python 3.12 and officially support 3.12 throughout the message stack :)

glatterf42:
The publish workflow doesn't run because https://github.com/iiasa/message-ix-models/actions/runs/8433520273 is neither starting (despite the job it's waiting for being finished) nor able to get cancelled. If it's okay with you, @khaeru, I'd change the required checks and merge this, anyway.

khaeru:
Could you please adjust the required statuses in the repo settings? These have different names now with these changes to the workflow. Otherwise good to go.

PR #151, opened 2024-01-29
Raise informative exception from ScenarioInfo.io_units()

khaeru:
Currently, if either the commodity or technology lacks units, the exception is internal to pint and could be hard to interpret:
```
File "/home/khaeru/vc/iiasa/models/message_ix_models/util/scenarioinfo.py", line 265, in io_units                                                                                                                                  
  return self.units_for("commodity", commodity) / self.units_for(                                                                                                                                                                  
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                
File "/home/khaeru/.venv/3.11/lib/python3.11/site-packages/pint/facets/plain/unit.py", line 159, in __truediv__
  return self._REGISTRY.Quantity(1 / other, self._units)
                                 ~~^~~~~~~
TypeError: unsupported operand type(s) for /: 'int' and 'NoneType'
```

## How to review

## PR checklist

- [x] Update doc/whatsnew.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -7,6 +7,7 @@ Next release
 - Add :ref:`tools-iea-web` for handling data from the International Energy Agency (IEA) Extended World Energy Balances (:issue:`25`, :pull:`75`).
 - Add :ref:`tools-wb` and :func:`.assign_income_groups` to assign MESSAGE regions to World Bank income groups (:pull:`144`).
 - Adjust :mod:`.report.compat` for genno version 1.22 (:issue:`141`, :pull:`142`).
+- Raise informative exception from :meth:`.ScenarioInfo.io_units` (:pull:`151`).
 
 v2023.11.24
 ===========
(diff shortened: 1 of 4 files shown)


khaeru:
FYI @glatterf42, I was able to debug this issue with a temporary commit that made a change like this:
```py
@pytest.fixture(scope="function")
def test_context(request, session_context):
    """A copy of :func:`session_context` scoped to one test function."""
    from message_ix_models.model.config import Config

    ctx = deepcopy(session_context)

    # Defaults are preserved on session_context and thus ctx
    assert session_context.model.regions == Config.regions

    # Ensure there is a report key
    ctx.setdefault("report", dict())

    yield ctx

    ctx.delete()

    # Defaults are preserved on session_context and thus ctx
    assert session_context.model.regions == Config.regions
```
This helped me see that the assertions failed _after_ tests which used the `mix_models_cli` fixture. This one was using the `session_context` fixture, and so changes made to that Context by various CLI commands would be copied within `test_context` and on to tests.

The last commit on this branch has the fix.

glatterf42 (approved):
Looks good to me, thanks for fixing this :)

PR #157, opened 2024-03-20
Create "Quick starting guide to running MESSAGEix-GLOBIOM" for occasional users

glatterf42:
This PR supersedes https://github.com/iiasa/message_data/pull/312 and thus fixes https://github.com/iiasa/message_data/issues/287 once we merge it. The main reason for migrating this PR is that we now have a snapshot of a global baseline model available at Zenodo, so there might be demand for having these instructions be public, too.

Thanks a lot to @jkikstra for writing up the original issue and to @LauWien for starting the original PR. 

After going through the original material, I think almost everything should be included in the new docs except for some Common Issue that @jkikstra reported:

> - a GAMS error the first time you are doing `scenario.solve()`:
  ```
  com.gams.api.GAMSException: com.gams.api.GAMSException: GAMS system directory is both not specified and not found from your environment!
  ```
> FIX: You have not yet set up your `GAMSDIR` and `Path` environment variables correctly. Set `GAMSDIR` for instance to "C:\GAMS\35; C:\GAMS\35\gbin" and add to `Path` "%GAMSDIR%"
 
> - a GAMS license error the first time you are doing `scenario.solve()`:
  ```
  ![image](https://user-images.githubusercontent.com/46967503/152387421-c16464bb-d985-4221-b1bc-8f8e0db78288.png)
  ```
> FIX: locate your "gamslice.txt" (if you don't have this, ask someone else about GAMS license files) and place it in your GAMS folder (e.g. "C:\GAMS\35"). After that, copy the text of that file to your clipboard, then open GAMS Studio, go to Help -->GAMS Licensing, which should then give a pop-up that will solve all of your problems.

> - a GAMS execution error relating to the MACRO iteration, e.g. 
  ```
  ixmp.model.base.ModelError: GAMS errored with return code 3:
      There was an execution error
  For details, see the terminal output above, plus:
  Listing   : C:\Users\kikstra\Miniconda3\envs\mdl\lib\site-packages\message_ix\model\MACRO_run.lst
  Input data: C:\Users\kikstra\Miniconda3\envs\mdl\lib\site-packages\message_ix\model\data\MsgData_mdle-run-jsk_ssp2_macro.gdx
  ```
> FIX: To figure out what's up, we can check the log file (`*.lst`). Open the `gamside` application and load the "MACRO_run.lst", and then search (Ctrl-F, then F3 for next) for "error" until you see something that helps you out.

The first one shouldn't happen if users follow the install instructions closely; the second one seems rather general, so should maybe make it into message_ix if we want to record it, but isn't directly related to install issues; the third one might also need to go to message_ix and is also only related to running the code without errors.

## How to review

@gidden, since you were interested in merging https://github.com/iiasa/message_data/pull/312, could you please check that no information in there is incorrect?

@yiyi1991, @daymontas1: since you have started recently, maybe you could take a look at the new docs and see if they would have been helpful for you. If you notice anything that seems to be missing, please leave a comment :)

FYI @khaeru, I might enable building the docs on RTD for this PR to make it easier to check the new ones. 

- Look at the following pages of the documentation:
  - [Installation](https://iiasa-energy-program-message-ix--157.com.readthedocs.build/projects/models/en/157/install.html)
  - [Quickstart](https://iiasa-energy-program-message-ix--157.com.readthedocs.build/projects/models/en/157/quickstart.html)

## PR checklist

- ~[ ] Add or expand tests; coverage checks both ✅~ Only updating docs.
- [ ] Update doc/whatsnew.

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -56,10 +56,6 @@ The following sets of extra dependencies are available; per the user guide linke
    Minimal requirements for the test suite.
 
 
-Known issues
-------------
-
-
 Check that installation was successful
 --------------------------------------
 
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -1,19 +1,77 @@
+.. _installation:
 Installation
 ************
 
+.. note:: :mod:`message_ix_models` requires :mod:`message_ix` to run. Please ensure your system has `their required dependencies <https://docs.messageix.org/en/stable/install.html#install-system-dependencies>`_ installed.
+
+:mod:`message_ix_models` is structured as a Python package and is published to the PyPI public code repository. Hence, there are two options for the installation:
+
+
 From PyPI
-=========
+---------
+
+This option is only recommended for users who do not wish to make any changes to the source code.
 
 1. Run::
 
-    $ pip install message-ix-models
+    $ pip install message-ix-models[<extra_dependencies>]
 
 
 From source
-===========
+-----------
 
-1. Clone the code.
-2. Run::
+Use this option if you intend to make changes to the source code. 
+We value your contributions via pull requests to `the main repository <https://github.com/iiasa/message-ix-models>`_. Please consider `contributing to development <https://docs.messageix.org/en/latest/contributing.html>`_ your changes.
 
-    $ pip install --editable .
+1. Fork the `the main repository <https://github.com/iiasa/message-ix-models>`_.
+   This will create a new repository ``<user>/message-ix-models``.
 
+2. Clone your fork; using the `Github Desktop <https://desktop.github.com>`_ client, or the command line::
+
+    $ git clone git@github.com:USER/message-ix-models.git
+
+3. Add the main repository as a remote git repository.
+   This will allow keeping up to date with changes there and importing tags, which also needs to be done for the install tests to succeed::
+
+    $ git remote add upstream git@github.com:iiasa/message-ix-models.git
+
+    $ git fetch upstream --tags
+
+4. Inside the ``message-ix-models`` folder, run::
+
+    $ pip install --editable .[<extra_dependencies>]
+
+
+Dependencies
+------------
+
+See :file:`pyproject.toml`.
+The following sets of extra dependencies are available; per the user guide linked above, they can be installed along with the mandatory dependencies by adding (for instance) ``extra_name`` to the package spec :program:`pip install message_data[extra_name]`.
+
+``docs``
+   Minimum requirements for building the docs.
+``report``
+   For running the :mod:`~.message_ix_models.report` functionality.
+``tests``
+   Minimal requirements for the test suite.
+
+
+Known issues
+------------
+
+
+Check that installation was successful
+--------------------------------------
+
+Verify that the version installed corresponds to the `latest release <https://github.com/iiasa/message-ix-models/releases>`_ by running the following commands on the command line::
+
+    # Show versions of message_ix, message-ix-models, and key dependencies
+    $ message-ix show-versions
+
+    # Show the list of modelling platforms that have been installed and the path to the database config file
+    # By default, just the local database should appear in the list
+    $ message-ix platform list
+    $ mix-models config show
+
+The above commands will work as of :mod:`message_ix` 3.0 and in subsequent versions.
+Please read through the output ``mix-models`` :mod:`~.message_ix_models.cli` command to understand the different CLI options and what you can do with them. You can also study its `documentation page <https://docs.messageix.org/projects/models/en/latest/cli.html>`_ online.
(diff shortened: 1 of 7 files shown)


glatterf42:
Curious as to why the tests fail to read the snapshot. Is there some kind of rate limit with Zenodo that we are hitting?

khaeru:
Overall this seems like a good start!

Things that jump out at a quick skim:
- Several sections of this appear to be simply copy-pasted from other docs, for instance the instructions on how to configure the connection details for a particular ixmp Platform.

  We should instead link directly to the relevant sections and only add context-specific information. We could also update or expand those in-place. Duplicated text means extra work to remember to keep both in sync in the future.
- In several places where such links are already used, they are written a ReST URL links to https://docs.messageix.org/… . These should be InterSphinx links instead. I'll comment one example.

> Curious as to why the tests fail to read the snapshot. Is there some kind of rate limit with Zenodo that we are hitting?

Any test (or CI step) that makes a network query is vulnerable to transient connection issues: DNS lookups, the server responding, completing the query, etc. Some of these are not about our code but about the availability and steadiness of Zenodo's services. When we try this enough times, a low failure probability can yield frequent single failures.

doc/install.rst
```diff
@@ -1,19 +1,73 @@
...
-    $ pip install message-ix-models
+    $ pip install message-ix-models[<extra_dependencies>]
 
 
 From source
-===========
+-----------
+
+Use this option if you intend to make changes to the source code. 
+We value your contributions via pull requests to `the main repository <https://github.com/iiasa/message-ix-models>`_. Please consider `contributing to development <https://docs.messageix.org/en/latest/contributing.html>`_ your changes.
```
khaeru, line 24: This is a direct URL link:
```rest
`contributing to development <https://docs.messageix.org/en/latest/contributing.html>`_
```

Wherever possible we use [intersphinx links](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html), which are more reliable:
```rest
:doc:`message-ix:contributing`
```

- The prefixes before the `:` can be viewed in conf.py: https://github.com/iiasa/message-ix-models/blob/583d3e90a8ad5a8afcd28f4d7a1683e883c4d9ee/doc/conf.py#L134-L150
- The [docs show](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#showing-all-links-of-an-intersphinx-mapping-file) how to retrieve and inspect the list of :doc:, :ref: etc. targets available from a given other project.

glatterf42: The command to retrieve the targets doesn't readily work for me. If I pass a directory to it, it gives me an `IsADirectoryError`; if I pass in a single file, I receive varying `ValueError`s because the inventory header is invalid; and the same is true if I pass in a URL to a page of our docs.
I guess I'll look up targets manually.

khaeru: Notice that in the example in the Sphinx docs, `…/objects.inv` is appended. This happens automatically during `sphinx-build`, but not for the helper command.

glatterf42: Right, it needs to be the `objects.inv` file that I can find in e.g. `ixmp/doc/_build/`.

yiyi1991:
This merge to #287 is super helpful. Thank you. 
Maybe we can also shortly mention the required GMAS license version (v29+ I believe) if it is not mentioned elsewhere.

glatterf42:
The docs build might not work yet because I already included some docs labels from message_ix that are only yet introduced with https://github.com/iiasa/message_ix/pull/805.

Apart from that, I rewrote parts of the new docs to avoid duplicate information, but there's actually not a lot of copy-paste in here to begin with, so it's still the same in large parts.

doc/install.rst
glatterf42: If I understand correctly, intersphinx tries to resolve an undefined reference like `:doc:'installation'` by going through all other linked sources, so I tried following the [official guide](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#explicitly-reference-external-objects) and linked external reference in this way: `:external+message-ix:doc:'installation'`.

glatterf42:
> Maybe we can also shortly mention the required GMAS license version (v29+ I believe) if it is not mentioned elsewhere.

We already do that: https://docs.messageix.org/en/stable/install.html

glatterf42:
Now, there are only the two links not working that rely on the new labels in message_ix. Even though our `intersphinx_mapping` refers to `"https://docs.messageix.org/en/latest/"` for `message-ix`, I suspect this is because our `requirements.txt` installs `message_ix==3.8.0`.

Also, there's an abundance of unrecognized/unresolvable references in other documents and it would likely improve our docs a lot if we fixed them, but this is a topic for another PR.

khaeru (approved):
I added a commit to the branch with some copyedits and additions.

LGTM once checks pass!

glatterf42:
Thanks :)

Issue #27, opened 2021-07-21
Integrate ‘new’ and ‘old’ reporting


khaeru:
There are (as of 2021-07) two ways of doing reporting/and postprocessing for MESSAGEix-GLOBIOM–family models:
1. `message_data.reporting`, based on `message_ix.reporting`, `ixmp.reporting`, and `genno`—a.k.a. ‘new’ reporting.
2. ‘Old’ reporting, mainly contained in `message_data.tools.post_processing.iamc_report_hackathon.py`

(1) was developed to supersede (2) but—for operational reasons—(2) remains in parallel use. This issue is to:
- Discuss *how* to integrate these (short-/medium- term) and eventually retire (2) (long-term).
- Organize various efforts via other issues/PRs to carry out this work.

**(to be expanded)**

khaeru:
Today (2021-07-21) we had a 30-minute meeting with @volker-krey @OFR-IIASA @behnam-zakeri @adrivinca @francescolovat @LauWien to discuss.

khaeru:
As part of the NAVIGATE project, in message_data, some of this glue code (2) has been worked out for the {B,M,T} model variants. This is not yet documented.

khaeru:
#134 is a partial/experimental effort to replicate one of the trickier parts of the legacy reporting that prevented us moving away from it in the NAVIGATE project for calculation of CO₂ emissions from transportation. It runs on the snapshot model and solution data.

Similar code could be written to replicate other parts of the legacy reporting output.

khaeru:
@Jihoon notes that, even though we now have `message_ix_models.report.legacy` here, users may in some cases want to use this with certain branches of `message_data`:
- With a project- or variant-specific files:
  - Configuration files.
  - "Tables" from a Python module.
- With changes to the core legacy reporting files.

It would be good to, at least, document a suggested approach for this.

Issue #139, opened 2023-11-24
Split test_snapshot.test_load to a separate job


khaeru:
Current output from the "pytest" CI workflow includes [this](https://github.com/iiasa/message-ix-models/actions/runs/6977054334/job/18986365358#step:9:2428):
```
246.42s call     message_ix_models/tests/model/test_snapshot.py::test_load[0]
13.88s call     message_ix_models/tests/test_workflow.py::test_make_click_command
11.73s call     message_ix_models/tests/tools/test_advance.py::test_get_advance_data
```

At 4 minutes, this is more than half of the total job run time. To speed up CI, split this to a separate job.

PR #159, opened 2024-03-25
Migrate legacy reporting

glatterf42:
This PR takes the current version of the "legacy reporting" from [message_data's ssp_dev branch](https://github.com/iiasa/message_data/tree/ssp_dev) and plugs it into `message_ix_models/report/legacy`. 

No changes were made to the code in order to preserve comparability to other versions of the legacy reporting that might exist in unknown locations. This is why the files in `message_ix_models/report/legacy` had to be specifically excluded from code improvement tools such as mypy and ruff.

Since we want to confirm that the legacy reporting runs with the [latest version of the global model baseline snapshot on zenodo](https://zenodo.org/records/10514052), I added information about this snapshot to `SOURCES` in `util/pooch`. Some tests apart from `tests/report/test_legacy::test_legacy_report` might not work due to this change. In particular, [ss_reporter](https://github.com/iiasa/message-ix-models/blob/9c6096aad719a2ee7c6e69c67d1d2d6fa8406db4/message_ix_models/tests/test_report.py#L253-L270)  and [unpack_snapshot_data](https://github.com/iiasa/message-ix-models/blob/9c6096aad719a2ee7c6e69c67d1d2d6fa8406db4/message_ix_models/testing/__init__.py#L436-L455) seem to only handle `snapshot-0` for now.
On my local system, running `test_load` almost crashed my computer trying to unpack all `snapshot-1` data. It resulted in a `Java heap space OutOfMemoryError`.
On a tangent here, `ss_reporter` is a name where I'd vividly wish for a change, e.g. to `simulated_solution_reporter`. All these shortened names are bad in themselves, but even more so when this creates references to literal nazi organizations.

I didn't dare running `test_legacy_report`. I'm not sure the current setup is the best with `load_snapshots` a session-scoped fixture parametrized for the snapshots and yielding scenarios with them, but I'm not sure what else to do. `test_legacy_report` definitely still needs to be adapted to cover all function parameters/options, if that's our goal. We should at least find one working combination of those and bake it into the tests, I think.

Further TODOs that can be expected: 

- [x] The units for the latest snapshot version might have changed; make sure the `legacy_report()` can handle the new units
- [x] The DB receiving the scenarios might need to be taught the hierarchy between the data contained in the scenarios

Before starting this PR, I started migrating the legacy reporting while cleaning it up at the same time. This is nowhere near completion, but the point I got to can be seen at https://github.com/iiasa/message-ix-models/tree/modernize/legacy-reporting.

- Partly addresses #4
- Partly addresses #27

## How to review

TBA

## PR checklist

- [x] Update doc/whatsnew.


khaeru:
Thanks for getting this started!

Quick thoughts on only a few items:

> This is why the files in `message_ix_models/report/legacy` had to be specifically excluded from code improvement tools such as mypy and ruff.

I am undecided on whether we should also exclude them from coverage measurement. The legacy reporting is very big; in terms of line count it outweighs the entire existing message-ix-models code base. This combined with the low (previously: zero) test coverage would mean a big hit for the coverage of the package overall.

> On my local system, running `test_load` almost crashed my computer trying to unpack all `snapshot-1` data. It resulted in a `Java heap space OutOfMemoryError`.

In theory, the unpacking of the data goes parameter-by-parameter from the Excel file into the Gzipped CSVs, and shouldn't hold more than one in memory at a time. We could try forcing the backend to clear its cache after each one.

Also note that the repo contains a pre-unpacked version of snapshot-0, which saves time on CI. I think this might get unwieldy if we do this for every snapshot, but perhaps could do that in this case.

If this gets to be a big set of tasks, we could do it in its own PR and then rebase this one.

> `simulated_solution_reporter`

😬 Good catch, 100% we should change this.

> We should at least find one working combination of those and bake it into the tests, I think.

Agreed, 1 and no more for this initial PR. These can later be expanded as (internal) users need to ensure stable behaviour for other combinations of settings (though, again, we should prefer to expand the features of the genno-based reporting to provide what is needed).

glatterf42:
The tests are taking so long that I'll probably temporarily disable all but one version, too. And we might want to consider moving these tests to some kind of nightly workflow instead of testing on every push to every PR.

khaeru:
> …we might want to consider moving these tests to some kind of nightly workflow instead of testing on every push to every PR.

Indeed, see #139

glatterf42:
On `main`, we currently don't have the `.xlsx` snapshot files committed. With their most recent location update, I hope we finally register the existing unpacked data correctly, at least for snapshot-0. snapshot-1 might take significantly longer still since it's calling `ixmp.Scenario.read_excel()` directly without checking existince of other files first.
If this is indeed true, we might want to use `message_ix_models.model.snapshot.read_excel()` for snapshot-1, too.
Also, we should consider removing the `.xlsx` files from our commit history again to keep that smaller (or are they kept from `main`'s history automatically if we delete them here before merging?).

Either way, the snapshot-0 data still contain "Equation and variable data" that snapshot-1 is still missing. They were added through [this commit](https://github.com/iiasa/message-ix-models/commit/a0e2c0fc0f7fe2238bf017b971720d557bd11506), which describes the command with which they were created.

glatterf42:
This PR is going to create a headache because we are again including a `demand.csv.gz` and `DEMAND.csv.gz` and this time, we need both of them. `report/legacy/pp_utils.py` relies on `DEMAND` being there, unless we rewrite it.

Unfortunately, this duplicate name is not what's causing the current test failure. Testing locally, the check `if name in parameters` in `model/snapshot::read_excel` (correctly, I think) eliminates `DEMAND` from the list of parameters it wants to add. 
So I'm not quite sure why `read_excel(add_units=True)` doesn't add this unit anymore.

glatterf42:
Regarding test coverage, I waited with a decision to see how large the final drop would be. Even with the existing test (which is a single call to the legacy `report()` function, admittedly), this would lead to almost 30% reduced coverage, so I'm opting to exclude the migrated code from coverage instead.

glatterf42:
The main issue about reduced test coverage seems to be that the functions in `model/snapshot.py` like `unpack()` and `read_excel()` don't count as being tested since they are only run as part of a fixture. I'm not sure how we could tell codecov that this is intentional -- or maybe it shouldn't be, after all: maybe fixtures should only be very reliable code because the other tests depend on this code reliably running fine.

khaeru:
Towards getting this merged, can we please reword or drop the following commits?
```
8ba1a389 remove debug output to study failing scenario
e8c2afe7 Add debug output to study failing scenario
```

I will meanwhile go through and leave some comments.

message_ix_models/data/report/legacy/fil_files/AFR-HFC_fac.fil
```diff
@@ -0,0 +1,61 @@
+Variable,1990,1995,2000,2005,2010,2020,2030,2040,2050,2060,2070,2080,2090,2100,2110
```
khaeru, line 1: I would favour putting these under report/legacy/fil_files, until we have some documentation of their provenance, structure, code for handling them, etc.

message_ix_models/report/legacy/__init__.py
```diff
@@ -0,0 +1 @@
+from .iamc_report_hackathon import report as legacy_report
```
khaeru, line 1: We could perhaps omit this, since the intended entry-point for users is `message_ix_models.report.report()`.

message_ix_models/testing/__init__.py
```diff
@@ -351,3 +353,72 @@ def not_ci(reason=None, action="skip"):
...
+
+
+def unpack_snapshot_data(context: Context, snapshot_id: int):
+    """Already-unpacked data for a snapshot.
+
+    This copies the .csv.gz files from message_ix_models/data/test/… to the directory
+    where they *would* be unpacked by .model.snapshot._unpack. This causes the code to
+    skip unpacking them, which can be very slow.
+    """
+    if snapshot_id != 0 and snapshot_id != 1:
```
khaeru, line 365:
```py
if snapshot_id not in (0, 1)
```

message_ix_models/testing/__init__.py
```diff
@@ -351,3 +353,72 @@ def not_ci(reason=None, action="skip"):
...
+        log.info(f"No unpacked data for snapshot {snapshot_id}")
+        return
+
+    dest = context.get_cache_path(
```
khaeru, line 369:
```python
parts = (f"snapshot-{snapshot_id}", "MESSAGEix-GLOBIOM_1.1-R11_no-policy_baseline")
dest = context.get_cache_path(*parts)
```
Then `parts` can be reused further down instead of repeating.

message_ix_models/testing/__init__.py
```diff
@@ -351,3 +353,72 @@ def not_ci(reason=None, action="skip"):
...
+    shutil.copytree(snapshot_data_path, dest, dirs_exist_ok=True)
+
+
+@pytest.fixture(
+    scope="session",
+    params=[
+        int(k.split("-")[1]) for k in util.pooch.SOURCE if k.startswith("snapshot")
+    ],
+)
+def load_snapshot(request, session_context, solved: bool = False):
```
khaeru, line 391: The pytest style (that we try to follow) seems to be that fixture names are nouns instead of verbs like ordinary function names: for example, `tmp_path` instead of `make_tmp_path`.

So here perhaps `loaded_snapshot`?

We can also type annotate the function:
```py
def load_snapshot(...) -> "Scenario":
```
or similar.

message_ix_models/testing/__init__.py
```diff
@@ -351,3 +353,72 @@ def not_ci(reason=None, action="skip"):
...
+    snapshot_id: int = request.param
+    assert snapshot_id is not None
+    unpack_snapshot_data(context=session_context, snapshot_id=snapshot_id)
+    model_name = "MESSAGEix-GLOBIOM_1.1_R11_no-policy"
+    scenario_name = f"baseline_v{snapshot_id}"
+    mp = session_context.get_platform()
+
+    try:
```
khaeru, line 399: Maybe a comment here:
```py
# The following code roughly parallels bare_res()
```
This could be a cue that if we update one, the other might also need updating.

We might also try to merge the two functions, but not in this PR.

message_ix_models/testing/__init__.py
glatterf42: The type hint apparently needs to be `Generator[message_ix.Scenario, None, None]` based on [this answer](https://stackoverflow.com/a/38423388/23037943) since we're using `yield` instead of `return` in the fixtures. Hope that's fine.

message_ix_models/data/report/legacy/fil_files/AFR-HFC_fac.fil
khaeru: Ah, sorry—I misread the hierarchy! I thought they were in a directory /fil_files/ under the same parent directory as /report/. Please disregard.

glatterf42:
I'm not sure what to write in the docs. If I understand correctly, we generally want people to use the genno-reporting, but since the legacy reporting is designed to work with the public snapshot, we should probably mention this somewhere. I consider this to be the last ToDo before merging the PR.

khaeru:
I wonder if, given the long runtime of the "pytest-snapshot" CI workflow (25 minutes, versus 4–12 minutes for the "pytest" workflow jobs), we should:
- condition the run on [particular file paths changed](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), and/or
- run only/also on a schedule trigger.

But again, let's make such changes in (a) follow-up PR(s).

PR #162, opened 2024-04-03
Miscellaneous improvements for 2024-W14

khaeru:
- Migrate handing code for 4 sources of exogenous data from `message_data`.
  - Global Energy Assessment project.
    - Add fuzzed data for testing.
  - Global Fuel Economy Initiative.
    - Add fuzzed data for testing.
  - IEA Energy Efficiency Indicators.
    - For this source, **no** fuzzed data are added for testing. This is because the source uses a complicated, idiosyncratic file format. Duplicating this format is out of scope for the current PR. The lines are thus excluded from test coverage.
    - For the 2023-11 edition of these data, the IEA website at https://www.iea.org/data-and-statistics/data-product/energy-efficiency-indicators provides a "Demo and availability file", which we could use for testing purposes.
    - → #164
  - SHAPE project.
    - Add fuzzed data for testing.
- Improve docs of all of the above.
- Add `path_fallback()`, generalized from a utility function used in message_data/MESSAGEix-Transport.

Housekeeping:
- Adjust MANIFEST.in after #159.
- Remove snapshot files also available from Zenodo; these were committed in #159.
- Use [nschloe/action-cached-lfs-checkout](https://github.com/nschloe/action-cached-lfs-checkout) to reduce Git LFS bandwidth usage.
  - As we accumulate more Git LFS files (particularly, test data) in this repo, we are consuming the 300 GB monthly quota more and more quickly. One reason is that we currently run 18 jobs (6 upstream versions × 3 OSes), so even 100 MB of Git LFS files means a 1.8 GB hit towards that quota
  - This added, third-party action seems to be a response to others encountering the same issue. It appears the most developed, but is not 100% up to date; see some pending PRs https://github.com/nschloe/action-cached-lfs-checkout/pulls
  - We could mitigate by creating a fork or using someone else's.
- Configure Codecov to allow coverage to drop by 0.5% > 0.46% = 90 lines. This drop will (currently) always appear on PRs, because the snapshot and legacy reporting tests are only run on the `main` branch and a schedule trigger.
- Tidy a lot of nitpicks in docs.

## How to review

## PR checklist

- [x] Update doc/whatsnew.


glatterf42 (approved):
Thanks for these manifold improvements :)

PR #163, opened 2024-04-05
Adapt to new WB data and API

glatterf42:
Some recent changes in the WB data and API have led to our tests crashing (unfortunately, I found no release notes or some such to link to except for our [failing tests](https://github.com/iiasa/message-ix-models/actions/runs/8565257861/job/23473060209)). This PR adapts the expected SHA of the data file and the API so that we can still retrieve the data we require.

## How to review

  does in the future.

## PR checklist

- [x] Update tests; coverage checks both ✅
- ~[ ] Add, expand, or update documentation.~ Just an internal update.
- ~[ ] Update doc/whatsnew.~ Just an internal update.

### DIFF
--- a/message_ix_models/tools/wb.py
+++ b/message_ix_models/tools/wb.py
@@ -69,7 +69,7 @@ def assign_income_groups(  # noqa: C901
     elif method == "population":
         # Retrieve WB_WDI data for SERIES=SP_POP_TOTAL (Population, total)
         dm = sdmx.Client("WB_WDI").data(
-            "WDI", key="A.SP_POP_TOTL.", params=dict(start_period=2020, end_period=2020)
+            "WDI", key="A.SP_POP_TOTL.", params=dict(startPeriod=2020, endPeriod=2020)
         )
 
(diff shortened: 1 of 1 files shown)


khaeru (approved):
Per discussion on Slack, added a temporary test mark that we can later remove. Good to go once CI completes.

I noticed that [the WB docs page](https://datahelpdesk.worldbank.org/knowledgebase/articles/906519) for these files (that we link from [our docs page](https://docs.messageix.org/projects/models/en/latest/api/tools.html#module-message_ix_models.tools.wb)) includes this text:

> You can also download the [current classification by income in XLSX format](https://datacatalogfiles.worldbank.org/ddh-published/0037712/DR0090755/CLASS.xlsx) and the [historical classification by income in XLSX format](https://datacatalogfiles.worldbank.org/ddh-published/0037712/DR0090754/OGHIST.xlsx).

The second file appears to *also* contain the current classification, so we could in the future adapt the code to use this file and provide classifications for different time periods. But this wouldn't avoid the fact that both files' hashes are likely to change each time they do an update.

PR #167, opened 2024-04-10
Convert Intratec data from Excel to CSV

measrainsey:
Convert Intratec data from Excel spreadsheet to CSV file

## How to review

## PR checklist

- [x] Update doc/whatsnew.


glatterf42 (approved):
Looks good to me :)

You might want to consider the name of your new file, though. When referring to regions, I think we typically use an upper-case `R`, so `R12` in this case. Also, if you foresee adding indices for other types of region distributions and possibly more files specific to them, you could also make `R12` a folder within `data/intratec/` and just have a file called `indices.csv` in there.

measrainsey:
> Looks good to me :)
> 
> You might want to consider the name of your new file, though. When referring to regions, I think we typically use an upper-case `R`, so `R12` in this case. Also, if you foresee adding indices for other types of region distributions and possibly more files specific to them, you could also make `R12` a folder within `data/intratec/` and just have a file called `indices.csv` in there.

Thanks @glatterf42! I implemented your advice and renamed the file/reorganized the folder structure.

I also added a line about the PR in `doc/whatsnew`.

khaeru:
@measrainsey, looking at where your code uses this data, I see this:

https://github.com/iiasa/message-ix-models/blob/8f39459fa2e978f80295d17feb0b052909dbeac2/message_ix_models/tools/costs/regional_differentiation.py#L560-L575

So it is not clear whether the file is for the R12 node codelist, or for R11 and gets adapted. From the values it seems the latter.

I suggest:
- Store the data for R11. Include the exact node names, like `R11_NAM`, in a column called "node".
- Call the existing function [`adapt_R11_R12`](https://docs.messageix.org/projects/models/en/latest/api/util.html#message_ix_models.util.node.adapt_R11_R12).

doc/whatsnew.rst
```diff
@@ -4,6 +4,7 @@ What's new
 Next release
 ============
 
+- Convert Intratec data for :doc:`/api/tools-costs` (:pull:`167`).
```
khaeru, line 7: "…from Excel to a simpler CSV format."

measrainsey:
> I suggest:
> 
>     * Store the data for R11. Include the exact node names, like `R11_NAM`, in a column called "node".
> 
>     * Call the existing function [`adapt_R11_R12`](https://docs.messageix.org/projects/models/en/latest/api/util.html#message_ix_models.util.node.adapt_R11_R12).

@khaeru: As per your suggestion, I changed the indices file to be for R11 instead (putting that file in an `R11` folder), and now use `adapt_R11_R12` when the R12 node is activated. (Just a note for myself on here that in order for `adapt_R11_R12` to work, the other column in the DataFrame that is passed through had to be called `value`.)

https://github.com/iiasa/message-ix-models/blob/35aa7630c66b1da7500ccd56e3101240995d66c0/message_ix_models/tools/costs/regional_differentiation.py#L539-L555

khaeru (approved):
Great, thanks for making those changes! From my POV also good to go.

Issue #169, opened 2024-04-11
Fix odd `tools.costs` projections for technologies with first years in the future


measrainsey:
There are a few technologies (mainly CCS) in the `tools.costs` module that have their first years after 2020 or 2021 (usually 2030). For some reason, their costs projections during between 2020 and the start year look kind of strange. See example below with `bio_istig_ccs`.

![bio_istig_ccs](https://github.com/iiasa/message-ix-models/assets/21201137/4aab2024-3842-4b2b-8bab-27f9fe2e2dc6)

I did a check and the reference region's cost at 2020 and 2030 (in this case) are the same. So, this shouldn't really have a big impact on the results if we're making sure to only add in the cost data from the start year onwards when updating the `inv_cost` or `fix_cost` parameters in MESSAGE.

The possible problems are that this (a) could be slightly affecting the results of the non-reference regions and (b) feels like bad form. I will add it to my list to fix this issue and to modify the tool to not project costs for technologies before their start year.

However, this does bring up the issue on how to deal with "base year" costs when a technology's start year is 2030, for example. Does the base year cost refer to start year cost or the cost in 2020 and therefore needed to adjust for start year?

glatterf42:
Please help me understand: do we already have something like "base year cost" in the model or is this an addition from your recent PR merger? 
If the former, I'd hope it's documented. If it's not documented or the latter, I'd draw the attention of specific experienced colleagues to this PR by mentioning them with `@`.

measrainsey:
@glatterf42 From what I understand, base year costs are not exactly in MESSAGEix itself, but base year costs are used when projecting investment costs and fixed operational and maintenance costs. This can be traced back to the older method of projecting costs [here](https://github.com/iiasa/message_data/tree/dev/data/model/investment_cost) in `message_data`. Specifically, the [NAM_technology_cost_input_20200507.xlsx file](https://github.com/iiasa/message_data/blob/dev/data/model/investment_cost/doc/NAM_technology_cost_input_20200507.xlsx) has details of how base year costs are derived for each technology (in this case, mostly energy technologies). 

The base year cost values in the spreadsheet linked above are used in the recent `tools.costs` PR merge, as an input data column called `base_year_reference_region_cost` (for example, in [this](https://github.com/iiasa/message-ix-models/blob/main/message_ix_models/data/costs/energy/tech_map_energy.csv) file). The data is read in by the [`get_raw_technology_mapping()`](get_raw_technology_mapping) function.

[...]

PR #186, opened 2024-04-24
Made changes to `tools.costs` related to base year and first year of modeling

measrainsey:
Made a number of changes to `tools.costs`, mostly relating to the modeling of base year, first model year, and first technology year

To summarize, the changes are:
- Fixed jumps in cost projections for technologies with first technology year that's after than the first model year (addresses issue #169) and remove projections for years that are before a technology's "first_technology_year"
- Changed the use of `base_year` to mean the year to start modeling cost changes
- Update cost assumptions for certain CCS technologies
- Changed the default fixed O&M reduction rate (`fom_rate`) to 0

Additional details relevant to each change found below.

## How to review

## PR checklist

- [x] Update doc/whatsnew.

### Fixed jumps in cost projections for technologies with later first technology year

Issue #169 described an issue with certain technologies that have a "first_technology_year" that's after the first model year (2020) behaving weirdly -- namely, the costs would jump up and then come back down to base year costs when it reaches its "first_technology_year". I've fixed this issue, so that the costs are projected just like other technologies (starting from the base year). Additionally, I've removed projections for years that are before the technology's "first_technology_year" completely. The "first_technology_year" refers to the first year a technology starts being deployed within MESSAGE. There's no other place this concept is used I think, so there's no other way to constrain it. So, I decided to filter out those years to avoid giving the model costs for years where the technology shouldn't even be in operation.

Using the same example as issue #169, here's how the investment cost projections for `bio_istig_ccs` looks now:
bio_istig_ccs

### Modified the usage of `base_year` in the module
Previously, the `base_year` was mainly just used to decide what year WEO data to use. Now, I've changed the module to use 2021 WEO data no matter what, and `base_year` means the year to start projecting cost changes/reductions. `base_year` does not have to equal the first model year (`y0`). If `base_year` is after `y0`, then the technology's costs is held constant at the base year cost until `y0`. Note that this means there will be no cost variance across scenarios until `base_year`, but there will still be regional differentiation. I've changed the default `base_year` to 2025, holding the cost across scenarios constant from `y0` (2020) to 2025 (so forcing no scenario variance from 2020 to 2025).

Example of this in action:

coal_ppl

### Updated cost assumptions for certain CCS technologies

@OFR-IIASA, @gidden, and others have noted that there are some cases with a CCS technology's costs would dip lower than its non-CCS variant. The reason for that is that based on the inputs into the cost module, since these are two different technologies, there is nothing stopping their assumptions from causing them to have completely different pathways. Using `syn_liq` and `syn_liq_ccs` as examples: for the regional differentation, `syn_liq` was being mapped to IGCC in the WEO, while `syn_liq_ccs` was mapped to IGCC with CCS. If WEO has different regional differentations for their two technologies (which they do), then that affects how our costs for `syn_liq` and `syn_liq_ccs`. Additionally, `syn_liq` and `syn_liq_ccs` were following different cost reduction pathways and narratives. The combinations of these issues caused the costs of `syn_liq_ccs` in some regions to dip lower than `syn_liq`, which doesn't make sense.

syn_liq_old

The list of problematic technologies found in the cost module where this phenomenon occurs are:

- h2_bio_ccs
- h2_coal_ccs
- liq_bio_ccs
- meth_coal_ccs
- meth_ng_ccs
- syn_liq_ccs

For the listed technologies above, I made the following changes to their data inputs:
- I changed the technology mapping so they are mapped to the same technology as their non-CCS variant
- I copied over the same scenario reduction rates and narratives from their non-CCS variant

Here is what `syn_liq_ccs` costs look like after the changes:

syn_liq_new

Longer term we are having discussions on if we want to move to component-based modeling of CCS components and adding that cost onto the original technology. 

### Changed the default fixed O&M reduction rate (`fom_rate`) to 0

@macflo8 took a look at `solar_pv_ppl` costs over time and noted that fixed O&M costs were increasing too rapidly, causing solar to be deployed at very slow rates. This is due to the default `fom_rate` in the module being 0.025, which is applied universally to all technologies. For now I've just changed this to zero, which means for each `year_vtg`, the costs remain the same across all `year_act`. Longer term I need to look at making this configurable for specific technologies (some have lower or higher values) -- perhaps this can just be an input similar to the base year reference region costs.


doc/whatsnew.rst
```diff
@@ -7,6 +7,12 @@ What's new
 v2024.4.22
 ==========
 
+- Made changes to :doc:`/api/tools-costs` (:pull:`186`).
```
glatterf42, line 10: These lines should go into the "next release" section since they were not actually included in v2024.4.22. To do that, please uncomment lines 4 and 5 (removing the `..`) and start a new list there :)

glatterf42: Especially if you anticipate more such changes, you could make this a whole subsection:
```
Changes to :doc:`/api/tools-costs` (:pull:`186`, <complete list later on>)
--------------------------------------------------------------------------

- Fix jumps in ...
```

doc/whatsnew.rst
```diff
@@ -7,6 +7,12 @@ What's new
...
+
+  - Fixed jumps in cost projections for technologies with first technology year that's after than the first model year.
```
glatterf42, line 12:
```suggestion
  - Fix jumps in cost projections for technologies with first technology year that's after the first model year.
```

Please always use present tense with these (below as well) :)

glatterf42 (commented):
Looks good to me, thanks :)
I'll approve once the docs-comments are addressed :)

doc/whatsnew.rst
measrainsey: Thanks @glatterf42 ! I've created a subsection for changes in the tools module (I think more changes could be forthcoming) and moved all that to the "next release" section.

glatterf42 (approved):
Looks good to me now :)

message_ix_models/tools/costs/projections.py
```diff
@@ -458,8 +468,21 @@ def create_message_outputs(
         )
         .astype(dtypes)
         .query("year_vtg in @config.Y")
+        .assign(first_technology_year=lambda x: x.first_technology_year.astype(float))
+        .assign(first_technology_year=lambda x: x.first_technology_year.astype(int))
```
khaeru, line 472: Is there any specific need for these two chained calls instead of directly?
```
.astype({"first_technology_year": int})
```

If there is, it would be good to record in a comment so people maintaining the code can understand why.

message_ix_models/tools/costs/projections.py
```diff
@@ -458,8 +468,21 @@ def create_message_outputs(
...
+        .query("year_vtg >= first_technology_year")
         .reset_index(drop=True)
-        .drop_duplicates()
+        .drop_duplicates()[
+            [
```
khaeru, line 476: AFAICT the intent of this selection is to drop the added column "first_technology_year". If so, then:
```
.drop("first_technology_year", axis=1)
```
…is more explicit and less noisy. Recall we made several adjustments like this in #99.

PR #187, opened 2024-04-25
Use 2023 WEO data in `tools.costs`

measrainsey:
Update costs module to move from using the 2022 WEO release to using the 2023 WEO data release

Main changes made:
- Upload 2023 data file
- Modify code to read in 2023 data release
- Use 2022 USD to 2001 USD conversion (instead of 2021 USD to 2001 USD)
- Use the new latest historical values (2022) for the regional differentation

Additionally, I've changed a few lines in the code to address the comments that were left on #186 which were not properly fixed in that PR.

## How to review

## PR checklist

- [x] Update doc/whatsnew.

### DIFF
--- a/message_ix_models/tools/costs/regional_differentiation.py
+++ b/message_ix_models/tools/costs/regional_differentiation.py
@@ -49,31 +49,31 @@ def get_weo_data() -> pd.DataFrame:
     # their respective sheet in the Excel file,
     # and the start row
     DICT_TECH_ROWS = {
-        "bioenergy_ccus": ["Renewables", 95],
-        "bioenergy_cofiring": ["Renewables", 75],
-        "bioenergy_large": ["Renewables", 65],
-        "bioenergy_medium_chp": ["Renewables", 85],
-        "ccgt": ["Gas", 5],
-        "ccgt_ccs": ["Fossil fuels equipped with CCUS", 25],
-        "ccgt_chp": ["Gas", 25],
-        "csp": ["Renewables", 105],
-        "fuel_cell": ["Gas", 35],
-        "gas_turbine": ["Gas", 15],
-        "geothermal": ["Renewables", 115],
-        "hydropower_large": ["Renewables", 45],
-        "hydropower_small": ["Renewables", 55],
-        "igcc": ["Coal", 35],
-        "igcc_ccs": ["Fossil fuels equipped with CCUS", 15],
-        "marine": ["Renewables", 125],
-        "nuclear": ["Nuclear", 5],
-        "pulverized_coal_ccs": ["Fossil fuels equipped with CCUS", 5],
-        "solarpv_buildings": ["Renewables", 15],
-        "solarpv_large": ["Renewables", 5],
-        "steam_coal_subcritical": ["Coal", 5],
-        "steam_coal_supercritical": ["Coal", 15],
-        "steam_coal_ultrasupercritical": ["Coal", 25],
-        "wind_offshore": ["Renewables", 35],
-        "wind_onshore": ["Renewables", 25],
+        "bioenergy_ccus": ["Renewables", 99],
+        "bioenergy_cofiring": ["Renewables", 79],
+        "bioenergy_large": ["Renewables", 69],
+        "bioenergy_medium_chp": ["Renewables", 89],
+        "ccgt": ["Gas", 9],
+        "ccgt_ccs": ["Fossil fuels equipped with CCUS", 29],
+        "ccgt_chp": ["Gas", 29],
+        "csp": ["Renewables", 109],
+        "fuel_cell": ["Gas", 39],
+        "gas_turbine": ["Gas", 19],
+        "geothermal": ["Renewables", 119],
+        "hydropower_large": ["Renewables", 49],
+        "hydropower_small": ["Renewables", 59],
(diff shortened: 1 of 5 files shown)


khaeru:
@measrainsey thanks for the PR and clean-ups!

For the moment I imagine no one will want to use the older WEO data. But suppose this:
- We (or others) publish work using the WEO 2023 data via this module.
- Development of `message-ix-models` carries on, with more updates and releases; to WEO 2024, WEO 2025, etc.
- At some point, someone would like to reproduce earlier work, e.g. work that used WEO 2023 per this PR.

The question is, how should they do that?
1. Simply revert to the older version, e.g. the 2024.4.22 release that had WEO 2022, or an upcoming release that has WEO 2023 per this PR.
   - This option could pose problems if there is no `pip freeze >requirements.txt` file, or if they are using other code that is compatible with future `message-ix-models` but not with 2024.4.22.
2. Provide a parameter or option (see for instance `.tools.iea.web`, or `.model.snapshot`) that allows the user to select the edition/version of input data used, and adjusts any other parameters to suit.
3. They can't; we will only support one version of WEO input data at a time, and if they want to use a different one (older or newer) they must figure it out on their own.

To be clear, I don't think *either* is necessary for this PR, but it would be good to consider which is desired for the future. @glatterf42 and I can then help implement.

measrainsey:
Hi @khaeru, thanks for your comment and feedback!
That's a good point you brought up -- option (2) would probably be best for the long term, both for other users and for ourselves in case we want to test how different releases of data affect our results.

measrainsey:
Circling back, would it be possible to review this PR as-is, and possibly work on a longer-term solution for using previous WEO data versions as a separate enhancement/functionality later on? I would like to get this merged sooner for the sake of the SSP update process. Thanks!

message_ix_models/tools/costs/regional_differentiation.py
```diff
@@ -49,49 +49,49 @@ def get_weo_data() -> pd.DataFrame:
...
 
     # Set file path for raw IEA WEO cost data
     file_path = package_data_path(
-        "iea", "WEO_2022_PG_Assumptions_STEPSandNZE_Scenario.xlsb"
+        "iea", "WEO_2023_PG_Assumptions_STEPSandNZE_Scenario.xlsx"
     )
 
     # Retrieve conversion factor
-    conversion_factor = registry("1.0 USD_2021").to("USD_2005").magnitude  # noqa: F841
+    conversion_factor = registry("1.0 USD_2022").to("USD_2005").magnitude  # noqa: F841
```
glatterf42, line 88:
```suggestion
    conversion_factor = registry("1.0 USD_2022").to("USD_2005").magnitude # if variable is needed later on
    _ = registry("1.0 USD_2022").to("USD_2005").magnitude # if not
```

glatterf42 (commented):
Certainly! As @khaeru mentioned, no need to tackle this issue now, though it would be good to create an issue tracking it.

The PR looks good to me, just one nit-pick: I think F841 errors regard variables that are assigned to but never used; these are easy to fix and don't ever need a `# noqa` statement. If you need the variable `conversion_factor`, you should be fine without the `# noqa`, and if you don't need it, you can use an `_` to clarify. After that, happy to approve :)

message_ix_models/tools/costs/regional_differentiation.py
measrainsey: Thanks @glatterf42 ! The `conversion_factor` is an interesting case where it _is_ used later on here:

https://github.com/iiasa/message-ix-models/blob/0b0ee50c11d1a3c7dfa3b8dfe94afd9afbf35973/message_ix_models/tools/costs/regional_differentiation.py#L125

But when used inside the string, the linting does not pick it up as being used. Without the `# noqa: F841`, when I save the file, the linter will automatically delete the `conversion_factor` variable line. So that's why the `# noqa: F841` is put there. Alternatively, I can change the variable name to `_conversion_factor` and use it in the string as `@_conversion_factor`, and I think that should work as well. 😄

glatterf42: Huh, I didn't know about such a use-case. However, reading up on it, I found [this note](https://pandas.pydata.org/pandas-docs/stable/user_guide/enhancingperf.html#expression-evaluation-via-eval): `DataFrame.eval()` is only recommended for `DataFrames` with more than 1e5 rows. Do you have any estimate about the number of rows in your `df`? If it's lower for sure, [`df.assign()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.assign.html#pandas.DataFrame.assign) might be more performant.

measrainsey: I've changed the line to use `df.assign()` instead, so now `conversion_factor` is explicitly used :)

https://github.com/iiasa/message-ix-models/blob/5b3a1f820579562f0b8ac2a0e7e5368dbff8aefb/message_ix_models/tools/costs/regional_differentiation.py#L125

However, I will note that I believe encounter this same use-case in other places (especially when I use `df.query()`).

glatterf42 (approved):
Looks good to me :)
Following our [merge recommendations](https://docs.messageix.org/en/latest/contributing.html#merge), please merge without rebase or squash :)

PR #130, opened 2023-10-13
Add MESSAGEix-Materials module

GamzeUnlu95:
This PR migrates .model.material (i.e. MESSAGEix-Materials) from the (private) message_data repo to this message-ix-models package, following the procedure described in https://github.com/iiasa/message-ix-models/pull/89.

## How to review

- Read the diff and note that the CI checks all pass.
- Run the build, solve, and report commands as described in the documentation. 
- Build the documentation and check if the page looks decent. 
- Ensure that changes/additions are self-documenting, i.e. that another
  developer (someone like the reviewer) will be able to understand what the code
  does in the future.

## PR checklist
- [ ] Continuous integration checks all ✅
- [ ] Add or expand tests; coverage checks both ✅
- [ ] Update doc/whatsnew.

PR #188, opened 2024-04-25
Migrate MESSAGEix-Materials updates

macflo8:
This PR migrates the remaining updates of MESSAGEix-Materials to message-ix-models and is planned to complement #130 and potentially supersede it eventually.

## Major model changes

Demand projections for all materials commodities now processed all in one [module](https://github.com/iiasa/message-ix-models/blob/migrate-materials_2024-W17/message_ix_models/model/material/material_demand/material_demand_calc.py)

Base year demand data is moved from code to separate yaml files ([example](https://github.com/iiasa/message-ix-models/blob/migrate-materials_2024-W17/message_ix_models/data/material/aluminum/demand_aluminum.yaml))

Implementation of SSP differentiated dimensions:
- Demand projections depend on population, GDP and elasticity assumptions of respective SSP
- Cost projections can be updated to SSP assumptions utilizing the [message-ix-models costs tool](https://github.com/iiasa/message-ix-models/blob/migrate-materials_2024-W17/doc/api/tools-costs.rst)

Refactored base year harmonization of industry energy consumption using latest [IEA data](https://www.iea.org/data-and-statistics/data-product/world-energy-statistics-and-balances)

## Steps run so far

The code has been migrated from the private repository with the procedure described in [the documentation](https://github.com/iiasa/message-ix-models/blob/main/doc/migrate.rst). 

## Further steps
- Write tests
- Compare with #130 and make sure all updates are collected in this branch

## How to review

- Run a specific code snippet or command and check the output.
- Build the documentation and look at a certain page.
  does in the future.

## PR checklist

~~- [ ] Continuous integration checks all ✅~~ Not possible due to missing tests
- ~- Add or expand tests; coverage checks both ✅~ See https://github.com/iiasa/message-ix-models/pull/189#issuecomment-2128709498; will be handled via #194 
- [x] Update doc/whatsnew.


glatterf42:
Thanks @macflo8, this will be very useful :)
Two immediate questions, though, about the files included: a lot of them are in `message_ix_models/data/material/version_control/` and I'm wondering: is `version_control` really a subtopic of `material`? If not, I suspect this directory was created to distinguish files under version control and those without, but all files on GitHub are necessarily under version control, so there would be no need to keep this. Something like `git mv message_ix_models/data/material/version_control/* message_ix_models/data/material/` might be enough to move all files up on level.
Second, some files are in `.../material/deprecated/`, which again makes me wonder: if these files are deprecated, do we really need to keep them around?

khaeru:
> Something like `git mv message_ix_models/data/material/version_control/* message_ix_models/data/material/` might be enough to move all files up on[e] level.

This could also be accomplished by giving a `--path-rename` argument to `git filter-repo` during the filtering process that moves files in that specific way.

khaeru:
At https://github.com/iiasa/message-ix-models/pull/189#issuecomment-2078748183 I mentioned the [commit message format](https://docs.messageix.org/en/latest/contributing.html#code-style) in our code style.

For this longer branch, it seems there are a lot more commits affected, so maybe the suggestion of manual changes during `git rebase -i` is not realistic. Please have a look at https://github.com/iiasa/message-ix-models/pull/107#issuecomment-2062609910, particularly the replacements.txt file used; I think extending this file with an appropriate regex would be an easy way to automatically rewrite e.g. "Feat(materials): make petrochemical elasticity ssp dependent" to "Make petrochemical elasticity SSP dependent".

Second, note that #186 has been merged recently. While it is *not* necessary to *constantly* rebase this branch and the one for #189, both should be based on the latest `main` at the future moment when we eventually actually do the two merges.

macflo8:
> Thanks @macflo8, this will be very useful :) Two immediate questions, though, about the files included: a lot of them are in `message_ix_models/data/material/version_control/` and I'm wondering: is `version_control` really a subtopic of `material`? If not, I suspect this directory was created to distinguish files under version control and those without, but all files on GitHub are necessarily under version control, so there would be no need to keep this. Something like `git mv message_ix_models/data/material/version_control/* message_ix_models/data/material/` might be enough to move all files up on level. Second, some files are in `.../material/deprecated/`, which again makes me wonder: if these files are deprecated, do we really need to keep them around?

Thanks for the comments @glatterf42

ad 1) Since we still rely on `xlsx` files for input data in MESSAGEix-Materials, keeping track of changes in these files is challenging since the changes are not displayed in human readable form. As a workaround I have added a [simple CLI command](https://github.com/iiasa/message-ix-models/blob/50d72245547731e179d6f8c17661e7c54934f2b4/message_ix_models/model/material/__init__.py#L793) to create a copies of all sheets in our `xlsx` files in `csv` format. Since we need to merge changes from some other branches like this [one](https://github.com/iiasa/message-ix-models/tree/update_steel), the `csv` files can help to identify the changes in the input files. So these files are not required for the MESSAGEix-Materials model and can be deleted before merging to `main`. This is not a very elegant solution, so if you have ideas for a better solution I am happy to discuss. My preference would be to move away from `xlsx` files eventually. But this would require a significant effort and a refactoring of input data is not at the top of our priority list at the moment..

ad 2) that is a good catch. Yes, I think they can be deleted. I see that they are also still present on the [original migrate branch](https://github.com/iiasa/message-ix-models/tree/migrate-materials-tidy/message_ix_models/data/material/deprecated). I will discuss it with @GamzeUnlu95.

glatterf42:
Thanks for the explanation, @macflo8 :)
For the input data file format, I don't have any good solution at the top of my head. I can however tell you that the [water](https://github.com/iiasa/message-ix-models/tree/main/message_ix_models/data/water/demands) and [buildings](https://github.com/iiasa/message-ix-buildings/tree/main/message_ix_buildings/sturm/data) modules use `csv` files for their input data, as far as I know. So maybe what you're doing is close to what you need to be compatible with them already: convert all excel files to `csv` once and adapt your code to accept that format rather than `xlsx`. I understand if this is not a top priority; there's no need to do this work in this PR (it might even be considered [scope creep](https://en.wikipedia.org/wiki/Scope_creep)). We can always open clean-up PRs with specific tasks to keep reviews manageable :)

glatterf42 (approved):
This is looking good! I think we can merge it with the understanding that #194 (in particular writing tests) should be addressed soon afterwards.

glatterf42:
This should be celebrated and mentioned in the next weekly MESSAGE meeting (at least :)

macflo8:
Yes definitely! 🥳 Many thanks to both of you @glatterf42 @khaeru for the help and guidance with the migration and everything else for this PR!

khaeru:
Very well done, and congrats!

PR #190, opened 2024-05-07
Change the default final year in `tools.costs` to 2110

measrainsey:
Update costs module to include 2110 in output by default

Currently the default `final_year` in `tools.costs` is 2100, which does not pair well with MESSAGEix's final year usally being 2110. Currently, users can change the `final_year` themselves as a parameter when creating their `Config` file, but given the use-base of this tool is mostly our internal team, I'm changing the default `final_year` to 2110.

Note that this does not mean costs are projected to 2110 now -- costs are still only projected to 2100. The module copies over 2100 projected costs for all years after 2100.

~~Also note that this PR is currently pointing to #187 -- I was thinking it would be best to wait for that PR to be merged first, then point this PR to main. But happy to revise as necessary.~~

Branch now has been rebased and points to main.

## How to review

## PR checklist

- [x] Update doc/whatsnew.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -4,13 +4,14 @@ What's new
 Next release
 ============
 
-Changes to :doc:`/api/tools-costs` (:pull:`186`, :pull:`187`)
+Changes to :doc:`/api/tools-costs` (:pull:`186`, :pull:`187`, :pull:`190`)
 -------------------------------------------------------------
   - Fix jumps in cost projections for technologies with first technology year that's after than the first model year.
   - Change the use of base_year to mean the year to start modeling cost changes.
   - Update cost assumptions for certain CCS technologies.
   - Change the default fixed O&M reduction rate to 0.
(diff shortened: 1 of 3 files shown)


doc/whatsnew.rst
```diff
@@ -4,13 +4,14 @@ What's new
 Next release
 ============
 
-Changes to :doc:`/api/tools-costs` (:pull:`186`, :pull:`187`)
+Changes to :doc:`/api/tools-costs` (:pull:`186`, :pull:`187`, :pull:`190`)
 -------------------------------------------------------------
```
glatterf42, line 8: Please expand to cover the whole line above.
Alternatively, you can collect the `:pull:` statements down below (e.g. after every bullet point) to avoid the need of adapting these two lines constantly.

glatterf42 (commented):
One comment in-line, otherwise looks good to me. We can finish #187 first, though :)

glatterf42 (approved):
Looks good to me once re-running test passes. Let's do another 'merge with commit' :)

PR #195, opened 2024-05-27
Create option in `tools.costs` to keep fixed O&M costs same across vintages

measrainsey:
Add option in costs model (`use_vintages`) which affect how fixed O&M costs are projected

Currently, fixed O&M costs (`fix_cost`) are projected in this way: for a certain `year_vtg`, the `fix_cost` is a percentage of investment cost (`inv_cost`). Then from the `year_vtg` to the end of the modeling horizon, the fixed O&M cost changes by a rate specified by `fom_rate` (as of present, the default `fom_rate` is 0, so there is no change).

However, this does not match with the current implementation in MESSAGEix. To match the MESSAGEix implementation, the `fix_cost` in a certain `year_act` should be the same across all `year_vtg`. As such, I've added an option to specify `use_vintage`. If `True` then the first projection method for `fix_cost` is used. If `False` (which is the default setting now), then `fix_cost` is calculated for each `year_act` instead as a percentage of the `inv_cost` in that year. 

Here is a figure showing the difference (Previous refers to `use_vintages=True` and Updated refers to `use_vintages=False`) using solar_pv_ppl as an example.

solar-fom-vintages

## How to review

## PR checklist

- [x] Update doc/whatsnew.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -17,7 +17,7 @@ Changes to :doc:`/api/tools-costs`
   - Change the default fixed O&M reduction rate to 0 (:pull:`186`).
   - Modify to use 2023 release of IEA WEO data and to use 2022 historic data for the base year (:pull:`187`).
   - Change the default final year to 2110 (:pull:`190`).
-  - Add option to not or not use vintages for fixed O&M costs (:pull:`195`).
+  - Add option to use or not use vintages for fixed O&M costs (:pull:`195`).
 
 v2024.4.22
 ==========
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -17,7 +17,7 @@ Changes to :doc:`/api/tools-costs`
   - Change the default fixed O&M reduction rate to 0 (:pull:`186`).
   - Modify to use 2023 release of IEA WEO data and to use 2022 historic data for the base year (:pull:`187`).
   - Change the default final year to 2110 (:pull:`190`).
-  - Add option to use or not use vintages for fixed O&M costs (:pull:`195`).
+  - Add :attr:`~.costs.Config.use_vintages` to control whether vintages are used in computing fixed O&M costs (:pull:`195`).  
 
 v2024.4.22
 ==========
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -8,9 +8,8 @@ Next release
 - Update :doc:`/material/index` (:pull:`201`).
 - Add :doc:`/project/edits` project code and documentation (:pull:`204`).
 - Reduce log verbosity of :func:`.apply_spec` (:pull:`202`).
+- Made fixes and updates to :doc:`/api/tools-costs` (:pull:`186`, :pull:`187`, :pull:`190`, :pull:`195`).
 
-Changes to :doc:`/api/tools-costs`
-----------------------------------
   - Fix jumps in cost projections for technologies with first technology year that's after than the first model year (:pull:`186`).
   - Change the use of base_year to mean the year to start modeling cost changes (:pull:`186`).
   - Update cost assumptions for certain CCS technologies (:pull:`186`).
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -17,6 +17,7 @@ Changes to :doc:`/api/tools-costs`
   - Change the default fixed O&M reduction rate to 0 (:pull:`186`).
   - Modify to use 2023 release of IEA WEO data and to use 2022 historic data for the base year (:pull:`187`).
   - Change the default final year to 2110 (:pull:`190`).
+  - Add option to not or not use vintages for fixed O&M costs (:pull:`195`).
 
 v2024.4.22
 ==========
--- a/message_ix_models/tools/costs/config.py
+++ b/message_ix_models/tools/costs/config.py
@@ -55,10 +55,11 @@ class Config:
     #: Model variant for which to project costs.
     module: Literal["energy", "materials"] = "energy"
 
-    #: Use vintages
-    #: If True, for each vintage, the fixed O&M costs will be calculated as a
+    #: Use vintages.
+    #:
+    #: - If True, for each vintage, the fixed O&M costs will be calculated as a
     #: ratio of the investment costs, that decreases by the rate of :attr:`fom_rate`.
-    #: If False, the fix_cost is the ratio of the investment cost for each year_act.
+    #: - If False, the fix_cost is the ratio of the investment cost for each year_act.
     #: In this case, the fix_cost is the same for all vintages of the same year_act.
     use_vintages: bool = False
 
--- a/message_ix_models/tools/costs/config.py
+++ b/message_ix_models/tools/costs/config.py
@@ -55,6 +55,13 @@ class Config:
     #: Model variant for which to project costs.
     module: Literal["energy", "materials"] = "energy"
 
+    #: Use vintages
+    #: If True, for each vintage, the fixed O&M costs will be calculated as a
+    #: ratio of the investment costs, that decreases by the rate of :attr:`fom_rate`.
+    #: If False, the fix_cost is the ratio of the investment cost for each year_act.
+    #: In this case, the fix_cost is the same for all vintages of the same year_act.
+    use_vintages: bool = False
+
     #: .. todo:: Document the meaning of this setting.
     pre_last_year_rate: float = 0.01
 
--- a/message_ix_models/tools/costs/config.py
+++ b/message_ix_models/tools/costs/config.py
@@ -57,9 +57,9 @@ class Config:
 
     #: Use vintages.
     #:
-    #: - If True, for each vintage, the fixed O&M costs will be calculated as a
+    #: If True, for each vintage, the fixed O&M costs will be calculated as a
     #: ratio of the investment costs, that decreases by the rate of :attr:`fom_rate`.
-    #: - If False, the fix_cost is the ratio of the investment cost for each year_act.
+    #: If False, the fix_cost is the ratio of the investment cost for each year_act.
     #: In this case, the fix_cost is the same for all vintages of the same year_act.
     use_vintages: bool = False
 
--- a/message_ix_models/tools/costs/projections.py
+++ b/message_ix_models/tools/costs/projections.py
@@ -393,27 +393,23 @@ def create_message_outputs(
         .rename(columns={"inv_cost": "inv_cost_2100", "fix_cost": "fix_cost_2100"})
     )
 
+    dims = [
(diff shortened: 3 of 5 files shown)


doc/whatsnew.rst
```diff
@@ -13,6 +13,7 @@ Changes to :doc:`/api/tools-costs`
   - Change the default fixed O&M reduction rate to 0 (:pull:`186`).
   - Modify to use 2023 release of IEA WEO data and to use 2022 historic data for the base year (:pull:`187`).
   - Change the default final year to 2110 (:pull:`190`).
+  - Add option to not or not use vintages for fixed O&M costs (:pull:`195`).
```
glatterf42, line 16:
```suggestion
  - Add option to use or not use vintages for fixed O&M costs (:pull:`195`).
```

message_ix_models/tools/costs/projections.py
```diff
@@ -477,67 +489,115 @@ def create_message_outputs(
...
-        .drop(columns=["key"])
-        .query("year_act >= year_vtg")
-        .assign(
-            val=lambda x: np.where(
-                x.year_vtg <= y_base,
-                np.where(
-                    x.year_act <= y_base,
-                    x.fix_cost,
+
+    if config.use_vintages is True:
```
glatterf42, line 493:
```suggestion
    if config.use_vintages:
```
Not a necessary change, but since this is a bool, `if config.use_vintages` will evaluate to `True` if and only if this setting is `True` and `False` iif it's `False`.

glatterf42 (approved):
Thanks, looks good to me :)
Some minor comments in-line :)

message_ix_models/tools/costs/projections.py
```diff
@@ -397,11 +397,23 @@ def create_message_outputs(
...
+                    "scenario",
+                    "message_technology",
+                    "first_technology_year",
+                    "region",
+                ],
             )
             .merge(
                 val_2100,
-                on=["scenario_version", "scenario", "message_technology", "region"],
+                on=[
```
khaeru, line 410: For these chained pandas calls, it is much more readable to define a list of dimensions/column names before the chain:
```py
dims = ["scenario_version", "scenario", ..., "region"]
```
and then reference it:
```py
.merge(val_2020, on=dims)
.merge(val_2100, on=dims)
```

message_ix_models/tools/costs/projections.py
```diff
@@ -477,67 +489,115 @@ def create_message_outputs(
...
+            .rename(
+                columns={
+                    "fix_cost": "value",
+                    "message_technology": "technology",
+                    "region": "node_loc",
+                }
+            )
+            .reindex(
+                [
+                    "scenario_version",
```
khaeru, line 579: Same here; saves the reader having to eyeball-diff to see that the list is the same as in the prior `if`-branch.

message_ix_models/tools/costs/projections.py
```diff
@@ -477,67 +489,115 @@ def create_message_outputs(
...
-        .query("year_act >= year_vtg")
-        .assign(
-            val=lambda x: np.where(
-                x.year_vtg <= y_base,
-                np.where(
-                    x.year_act <= y_base,
-                    x.fix_cost,
+
+    if config.use_vintages is True:
+        fom = (
```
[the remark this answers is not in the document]
khaeru: As far as I can tell after squinting at it for a few minutes, the only difference between the two branches of the `if:` block is that one contains `.assign(val=lambda x…)`, and the other does not.

In this case, instead of duplicating a lot of code, I'd suggest to use a temporary/internal function that passes-through if nothing is to be done, something like:
```python

def _compute_value(df: pd.DataFrame) -> pd.Series:
    if config.use_vintages is False:
         return df["fix_cost"]  # Pass through

    return np.where(...)

fom = (
   df.merge.drop(columns=["inv_cost"])
   .assign(key=1)
   ...
   .assign(value=_compute_value, unit="USD/kWa")
   .rename(columns=…)
   ...
)
```

doc/whatsnew.rst
```diff
@@ -13,6 +13,7 @@ Changes to :doc:`/api/tools-costs`
   - Change the default fixed O&M reduction rate to 0 (:pull:`186`).
   - Modify to use 2023 release of IEA WEO data and to use 2022 historic data for the base year (:pull:`187`).
   - Change the default final year to 2110 (:pull:`190`).
+  - Add option to use or not use vintages for fixed O&M costs (:pull:`195`).
```
khaeru, line 16: It's possible to link directly to the new Config class attribute:
```rst
- Add :attr:`~.costs.Config.use_vintages` to control whether vintages are used in computing fixed O&M costs (:pull:`195`).
```

This lets the changelog entry be brief, and the documentation of the item *per se* to contain full detail.

message_ix_models/tools/costs/config.py
```diff
@@ -55,6 +55,13 @@ class Config:
     #: Model variant for which to project costs.
     module: Literal["energy", "materials"] = "energy"
 
+    #: Use vintages
+    #: If True, for each vintage, the fixed O&M costs will be calculated as a
+    #: ratio of the investment costs, that decreases by the rate of :attr:`fom_rate`.
+    #: If False, the fix_cost is the ratio of the investment cost for each year_act.
+    #: In this case, the fix_cost is the same for all vintages of the same year_act.
+    use_vintages: bool = False
```
khaeru, line 63: Nice, clean addition! This is one way using a Config class helps clarify changes to code.

Please check how the docstring renders. I think Sphinx will rewrap the first two lines so they look like "Use vintages If True" (with no period).

message_ix_models/tools/costs/projections.py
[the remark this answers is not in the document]
measrainsey: Hi! Thanks all for both your comments. 

You're correct @khaeru that there's not a big difference between the two. However, the main differences are:

1. The `assign(val=lambda x...)` like you mentioned and
2. The `merge` command takes a different input. If `use_vintage` is True, then it is:
```
            .merge(
                pd.DataFrame(data={"year_act": config.seq_years}).assign(key=1),
                on="key",
            )
```
If it is False, then the merge command is:
```
            .merge(
                pd.DataFrame(data={"year_vtg": config.seq_years}).assign(key=1),
                on="key",
            )
```
So it is a matter of whether the dataframes are merged on `year_vtg` or `year_act`. Given this, would you still recommend implementing a function such as `_compute_value()`?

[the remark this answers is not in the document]
khaeru: Per (2), you can:
```python
# Construct a variable according the config setting:

to_merge = pd.DataFrame(
    {"year_act" if config.use_vintage else "year_vtg": config.seq_years}
).assign=key=1

# And then in the chained call, reference whatever the variable contains:
    ...
    .merge(to_merge, on="key")
    ...
```

doc/whatsnew.rst
measrainsey: Updated, thanks for the suggestion!

khaeru:
A minor request: can we please use nested lists rather than a sub-heading in doc/whatsnew? See for example https://github.com/iiasa/message-ix-models/blob/dad8f8a5cbdacb85ae8881e267982dad944b8c70/doc/whatsnew.rst?plain=1#L113-L127

Generally we reserve use of sub-headings to separate "All changes" from an occasional section "Migration notes" when we need to give clear instructions to users on how to adapt to changed code.

message_ix_models/tools/costs/config.py
measrainsey: Thanks! 

I have edited to account for the docstring rendering. 🙂

message_ix_models/tools/costs/projections.py
measrainsey: Thanks! Have updated as such:

https://github.com/iiasa/message-ix-models/blob/08ccfd6504c8be03e63718889eaa102677d7d6b3/message_ix_models/tools/costs/projections.py#L396-L413

message_ix_models/tools/costs/projections.py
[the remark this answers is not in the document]
measrainsey: Sorry for the delay - thank you for your suggestions! I have edited the code to be a bit more condensed and not have to be duplicated as much:

https://github.com/iiasa/message-ix-models/blob/b3da5aa67c62c07f9dbe847a8ccc555d3b620f61/message_ix_models/tools/costs/projections.py#L489-L555

measrainsey:
> A minor request: can we please use nested lists rather than a sub-heading in doc/whatsnew? See for example
>
> Generally we reserve use of sub-headings to separate "All changes" from an occasional section "Migration notes" when we need to give clear instructions to users on how to adapt to changed code.

I hope I correctly understood what you meant, as I have now moved the changes related to the costs tool from a subheading to a list :)

https://github.com/iiasa/message-ix-models/blob/b3da5aa67c62c07f9dbe847a8ccc555d3b620f61/doc/whatsnew.rst?plain=1#L11-L19

message_ix_models/tools/costs/projections.py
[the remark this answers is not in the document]
khaeru: Thanks—this makes the code easier to review and, in the future, easier to maintain.

message_ix_models/tools/costs/projections.py
```diff
@@ -393,15 +393,23 @@ def create_message_outputs(
...
+        "first_technology_year",
+        "region",
+    ]
+
     df_merge = (
         (
             df_prod.merge(
                 val_2020,
-                on=["scenario_version", "scenario", "message_technology", "region"],
+                on=dims,
```
khaeru, line 408: I think if we remove the trailing comma, `black` or `ruff` will automatically reflow this method call into a single line. I will try to push a commit that does this.

khaeru (approved):
@measrainsey thanks for this clean and well-documented improvement. I will push one small code style change per the comment below, then merge.

Per:

> - [x] Add or expand tests; coverage checks both ✅

…it seems the tests are not actually changed or expanded to test the new feature. That would be desirable: even though the existing tests pass, and the checks show “codecov/patch — 77.7% of diff hit (target 52.3%)”, it would be good to confirm that this code works as you intend; that way we can guard against further improvements disrupting the intended behaviour.

However, I won't make that a blocked for this PR, only ask that you consider adding it to a subsequent one.

If tests are not updated, we usually write the following in the PR description:

```
```

which renders as:

> - [x] ~Add or expand tests;~ coverage checks both ✅

…to signal what was done.

khaeru:
The CI failures here are all due to non-responses from the World Bank's SDMX API, which I suspect are due to [this morning's global CrowdStrike bug/error](https://en.wikipedia.org/wiki/July_2024_global_cyber_outages).

I'll wait a few hours and re-run.

Issue #198, opened 2024-06-10
Further clean up material module


glatterf42:
Locally, I see a DeprecationWarning about this line:

https://github.com/iiasa/message-ix-models/blob/c6285cfdf5080f521a17020f29ffb3851677078e/message_ix_models/model/material/build.py#L5

The change required here is simple: use `from ixmp.util import ...` instead.

Issue #194, opened 2024-05-24
Test .model.material


macflo8:
This is a tracking issue for follow-up items from https://github.com/iiasa/message-ix-models/pull/188 and #189 . Create 1 or more PRs and (if necessary to keep discussion organized) issues to address:

- Add tests to raise coverage to 100% (@macflo8, @GamzeUnlu95)
- #192
- #193 
- Extend documentation of materials code (@macflo8, @GamzeUnlu95)
- Rebase https://github.com/iiasa/message-ix-models/tree/migrate-materials and decide how to move forward with #130

khaeru:
Another item to add: **packaging**
- Looking at [this workflow run](https://github.com/iiasa/message-ix-models/actions/runs/9222239094/job/25372993290#step:6:1527), we see a number of `.xlsx` files in /data/material/ are included in the build .whl and .tar.gz distributions.
- For the `.water` module and others, we have taken the approach of *excluding* these files from the distribution, and providing a function/CLI command to fetch these files from GitHub. See #137 (mixed with many other changes, sorry). See also [Data, metadata, and configuration](https://docs.messageix.org/projects/models/en/latest/data.html) in the docs.
- With that approach, the installation of message-ix-models remains small, and then users can fetch the data that they want to use. We don't force e.g. `.water`-focused users to download `.material` data, or vice versa.
- I would like to do this before we make a new release of message-ix-models that includes these materials `.xlsx` files.
- This is more or less orthogonal to #193, i.e. it applies whether the data files are CSV or Excel or any other format.

Issue #25, opened 2021-07-14
Retrieve and process IEA extended world energy balances


khaeru:
@volker-krey notes that the IEA Extended World Energy Balances are available via the OECD iLibrary:

- https://www.oecd-ilibrary.org/energy/data/iea-world-energy-statistics-and-balances_enestats-data-en
- These include the extended world energy balances, for annual editions from 2020 back to 2015.
- For the 2020 edition, the size is 750 MB compressed, 22 GB uncompressed, as ZIPped CSV.

`message-ix-models` should be enhanced to:
- Retrieve these files.
- Subset or filter the data where possible, to reduce size.
- Cache the results.
- Prepare model-ready inputs based on the data.

khaeru:
@OFR-IIASA reports that partial/prototype code for this is on a private/local branch somewhere.

khaeru:
Some thoughts about implementation:
- `genno` can be used, just as it's used in the pre-model data prep for MESSAGEix-Transport. In particular, it has features to cache intermediate results at different stages along a processing pipeline, and we can split the processing of different quantities from the full data set so that they can be executed separately.
- Because of the size of the data, performance will be a concern. Our code likely won't be able to have the entire raw data in memory at once, but will instead need to stream or split the processing.
- In order to test this code, we'd need a manually-prepared subset or generator of data in the same format.

phackstock:
@khaeru just a few quick questions about the caching part of the requirements since you mentioned that`genno` has this functionality. I implemented pretty much the exact same thing the other day using [`joblib.Memory`](https://joblib.readthedocs.io/en/latest/generated/joblib.Memory.html). It also hashes the input values of a function and saves the result as a pickle file.

I just skimmed the code of the caching part of `genno` and I'm not sure if two features which we might want and are done by my implementation are done by `genno`:

1. joblib.Memory doesn't only cache the input values but also the function code itself. This way if your function code change but the input stays the same it won't be tricked into wrongly thinking that it has the results cached already. 
[...]

PR #201, opened 2024-06-21
Fix discovered issues and improve MESSAGEix-Materials

macflo8:
- Removes `message_data` dependencies (some `.yaml` and `.csv` input files).
- Fixes incorrect imports of previously migrated code in `utils/compat/message_data`.
- Removes deprecated/unused variables and code sections.
- Adresses #198.
- Implements data packaging mentioned in https://github.com/iiasa/message-ix-models/issues/194#issuecomment-2129269295.
- Hide some debugging and project specific CLI commands of `material-ix` group
- Moves from using untested helpers in  `utils/compat/message_data` to using tested equivalents of `message-ix-models`
- Removes `click` options from `material-ix` commands that are already implemented in `message_ix_models/cli.py`
- Stores csv files in data/material as text (previously stored with git LFS)
- General updates to documentation

**Details:**
Building MESSAGEix-Materials on an existing MESSAGEix-GLOBIOM scenario still requires a few external data file that have previously been stored in a private repository. The respective files are migrated to `message-ix-models/data` and data paths in the modules updated accordingly in this PR. The files located in `util/compat/message_data` migrated in the course of #188 contain some incorrect import statements, which are fixed in this PR.

The following files are needed in the build and were therefore added:

- [SP_UE_dyn_input.xlsx](https://github.com/iiasa/message-ix-models/pull/201/files#diff-12088526ad75821ea474a07500ffa9edb33e7d586dc9e98d5b27f3cd87a1721c)
- [UE_dyn_input_all.xlsx](https://github.com/iiasa/message-ix-models/pull/201/files#diff-b4cb09592b6c71aca7e13b2bb3a5d29d8f5b5c390adddc4ee688b4523b14d5da)
- [IEA_mappings.csv](https://github.com/iiasa/message-ix-models/pull/201/files#diff-210c6fb5a6f78b9c9c41c1c654717cceab9c7660a1efe8fd9a1e179a28b70807)
- [IEA_mappings_chemicals.csv](https://github.com/iiasa/message-ix-models/pull/201/files#diff-bc6466b997552e13e17c48ff77d54eedd1c5a16350cb3b0fc7d467c6ad5452ee)
- [IEA_mappings_industry.csv](https://github.com/iiasa/message-ix-models/pull/201/files#diff-6e2f944c2a73d3ce7d21051f08220b0b0702b7d222369332109eaa242a8110c4)

The build however still requires proprietary data from [IEA Extended Energy Balances](https://www.iea.org/data-and-statistics/data-product/world-energy-statistics-and-balances) to calibrate historical industry activity. The path to the required file can be specified with a newly added CLI option `--iea_data_path` in the `build` command.

All input data for MESSAGEix-Materials is now packaged in a `.tar.gz` file, that can be fetched with `mix-models fetch MESSAGEix-Materials`.

When executing `mix-models material-ix --help` only the basic comments related to building/solving/reporting are shown

## How to review

Run build/solve/report with `material-ix` CLI commands and confirm correct execution of each task.

## PR checklist

- ~~[ ] Continuous integration checks all ✅~~ Not possible due to missing tests
- ~~[ ] Add or expand tests; coverage checks both ✅~~ will be handled via https://github.com/iiasa/message-ix-models/issues/194
- [x] Update doc/whatsnew.


doc/material/index.rst
```diff
@@ -4,8 +4,9 @@ MESSAGEix-Materials
 Description
 ===========
 
-This module adds material stock and flow accounting in MESSAGEix-GLOBIOM. The implementation currently includes four key energy/emission-intensive
-material industries: Iron&Steel, Aluminum, Cement, and Chemicals.
+This module adds material stock and flow accounting in MESSAGEix-GLOBIOM. The
```
khaeru, line 7: Per the [code style](https://docs.messageix.org/en/latest/contributing.html#documentation), please:
> start each sentence on a new line, and do not hard-wrap within sentences.

message_ix_models/data/node/R12_SSP_V1.yaml
```diff
@@ -0,0 +1,84 @@
...
+  name: World
+  description: R12 regions
+
+R12_AFR:
+  parent: World
+  name: Sub-Saharan Africa
+  description: >-
+    Does not include Saint	Helena,	Ascension	and	Tristan	da	Cunha. 
+    In France-> Mayotte, Reunion data is present post 2010.
+    IEA MAP:  AFRICATOT - (DZA+EGY+LBY+MAR+SDN+SSD+TUN)->IIASA_AFRICA
```
khaeru, line 21: It's unclear what this description line and the following line mean, or how they are actually/intended to be used in the code:
- 'AFRICATOT' (for instance) is not defined anywhere.
- 'IIASA_AFRICA' (for instance) is not defined anywhere.
- This added file is not mentioned in doc/pkg-data/node.rst.
- The header comment seems to be exactly the same as the one in R12.yaml.

message_ix_models/model/material/__init__.py
```diff
@@ -657,7 +656,7 @@ def add_data_2(scenario, dry_run=False):
     log.info("done")
 
 
-@cli.command("modify_cost")
+@cli.command("modify_cost", hidden=True)
```
khaeru, line 659: In general, we prefer hyphens to underscores for CLI command and option names and values, e.g. `modify-cost`.

message_ix_models/model/material/__init__.py
```diff
@@ -805,3 +804,50 @@ def make_xls_input_vc_able(context, files):
         #     print(path, fname)
         # message_ix_models.model.material.util.excel_to_csv(files)
     return
+
+
+@cli.command("test_calib", hidden=True)
+@click.option("--scenario_name", default="NoPolicy")
```
khaeru, line 810: The top-level `mix-models` CLI has `--model` and `--scenario` options. Values for these are automatically handled and stored on the Context.

Rather than duplicate those, check the `context` and—if they are not set—fill in a default within the body of the command function.

message_ix_models/model/material/data_methanol_new.py
```diff
@@ -184,7 +184,7 @@ def broadcast_reduced_df(df, par_name):
 
         for col in yr_col_inp:
             yr_cols_codes[col] = literal_eval(df_bc_node[col].values[0])
-            broadcast_years(df_bc_node, yr_col_out, yr_cols_codes, col)
+            df_bc_node = broadcast_years(df_bc_node, yr_col_out, yr_cols_codes, col)
```
khaeru, line 187: The function called here has no tests or docstring. Does it do something that can't be done with [`message_ix_models.util.broadcast()`](https://docs.messageix.org/projects/models/en/latest/api/util.html#message_ix_models.util.broadcast)? If so, what?

khaeru (commented):
I added some comments inline.

Some other points:
- Files like message_ix_models/data/material/iea_mappings/*.csv:
  - should be stored as text, not using Git LFS.
  - contain the same phrase "iea_mappings" in their path and filename; this could be simplified.
  - do not appear to be mentioned anywhere in the documentation.
- `message_data.tools.utilities.get_optimization_years()` does the same thing as [`ScenarioInfo.Y`](https://docs.messageix.org/projects/models/en/latest/api/util.html#message_ix_models.util.scenarioinfo.ScenarioInfo.Y). The latter is computed only once, and has tests and documentation. The former is not tested, and hits the database every time it is called. I would prefer that we take the opportunity, when moving code to `message_ix_models`, to use the better-maintained solution. Similar applies to `get_historical_years()`.

message_ix_models/data/node/R12_SSP_V1.yaml
macflo8: This file was shared by @SiddharthJoshi-Git together with a copy of the preprocessed IEA EWEB 2023. It is used to group and aggregate IEA EWEB data to R12 regions for calibration purposes in MESSAGEix-Materials. Since the file is similar to R12.yaml with main differences in the `child` values, does it make more sense to merge it into the existing R12.yaml? Otherwise we can also rename the file since R12_SSP_V1.yaml is maybe not very indicative.

khaeru: > It is used to group and aggregate IEA EWEB data to R12 regions for calibration purposes in MESSAGEix-Materials.

Can you point to where specifically this happens?

macflo8: All of this happens in [material/data_util.py](https://github.com/iiasa/message-ix-models/blob/7c8cc4c731e8bfec19c1e3a8fca9f7489629fb42/message_ix_models/model/material/data_util.py). The top-level functions are `modify_industry_demand()` and `add_new_ind_hist_act()`, which are called in `model.material.build()`. They read a parquet file containing the EWEB data, map it to R12 and adjust existing `demand` and `historical_activity` parametrization. At the moment there is still some hardcoded parts in the whole procedure and inefficient parts (e.g. reading the parquet file more then once instead of passing it to the respective function).

khaeru: Okay, I see. I recall this was alluded to in a MESSAGE meeting recently—that some code, somewhere, is being used to process the IEA EWEB into a Parquet file, including some spatial manipulations that are not in [`.tools.iea.web`](https://docs.messageix.org/projects/models/en/latest/api/data-sources.html#extended-world-energy-balances-tools-iea-web). Not having seen that code, I can't guess at what it contains, but from the code in `message_ix_models.model.material`/this branch, it seems some of these labels like "AFRICATOT" are introduced or used by it. 

There are several ways in which this appears to duplicate existing `message_ix_models` code that has tests and docs and is more concise. We need to be careful to avoid this. However, to avoid expanding scope too much, I won't insist that they be removed/reworked this PR, but will mention them as further TODOs in #194 or subsidiary to it.

As for what to do in this PR:

Per your link, I see the file is used in two calls to this function https://github.com/iiasa/message-ix-models/blob/36fc249b7991d99b138d371cd8904a5356ac6851/message_ix_models/model/material/data_util.py#L921-L950

I am guessing here because of the lack of tests, but it appears most of the contents of `R12_SSP_V1.yaml` are not used. Rather, only the `child:` field is used to populate a "REGION" column in a data frame that has a "COUNTRY" column. Is that correct?

If so, this can be simplified a lot using existing features. For example, see: https://github.com/iiasa/message-ix-models/blob/dad8f8a5cbdacb85ae8881e267982dad944b8c70/message_ix_models/data/node/R12.yaml#L19
I suspect we can add another ‘annotation’ named, e.g. `material-iea-eweb-region` and containing these values like "CHINAREG". Then the file can be read with [`get_codes()`](https://docs.messageix.org/projects/models/en/latest/api/model.html#message_ix_models.model.structure.get_codes), and [`MappingAdapter`](https://docs.messageix.org/projects/models/en/latest/api/util.html#message_ix_models.util.MappingAdapter) can be used to do the relabeling.

I'd be happy to push some commits to make these changes; although again it will be difficult for me to ensure that I preserve behaviour without tests.

macflo8: Yes I was already wondering how we can harmonize the use of IEA EWEB across the team when I saw the feature being added to `message-ix-models` #25. I agree that this is out of scope for this PR. So we should continue this discussion in a separate issue I guess. The only issue with moving to `tools.iea.web` immediately that I see, is the missing pre-processing that is already applied in the parquet file. We would need to discuss the details with @SiddharthJoshi-Git.

Regarding the use of the `R12_SSP_V1.yaml`: Yes the mapping procedure only uses the `child` field. Adding a iea-eweb field to the original `R12.yaml` sounds like a good idea 👍 Feel free to push the commits that make use of `get_codes()` etc. I will try to write some simple tests today.

khaeru: I've just pushed the modifications, including one added test. Please have a look and try them out with the proprietary data file.

macflo8: Thanks, I manually tested with the private data file now. It seems that parametrization is identical except for R12_EEU. The private data file contains entries, where COUNTRY == "YUG", until 2021. This means there is a double counting of all countries that formerly belonged to Yugoslavia. According to the IEA EWEB documentation, IEA reports Yugoslavia under two codes: YUGOND and MYUGO. But I don't know how YUG data was derived in the private data file. So we need to either revisit the pre-processing of the EWEB data or modify the mapping in R12.yaml.

khaeru: Thanks for doing the check. The annotations I added to R12.yaml include:
```yaml
  material-region: ["*", SCG, YUG, KOSOVO]
```

The file added on the branch has:
```yaml
  child: [ALB, BGR, BIH, CZE, EST, HRV, HUN, LTU, LVA, MKD, MNE, POL, ROU, SRB, SVK, SVN, KOSOVO]
```
Curiously the latter does not include any of YUG, YUGOND, or MYUGO, which really makes me scratch my head.

But anyway, with the goal of identical behaviour, in the former, "*" refers to the list of children for the same node, which I see already includes "YUG". So you could try to edit to:
```yaml
  material-region: ["*", SCG, KOSOVO]
```
(don't duplicate, although with the way the dict is constructed, I don't think this will make any change)

or, most verbose:
```yaml
  material-region: [ALB, BGR, BIH, CZE, EST, HRV, HUN, LTU, LVA, MKD, MNE, POL, ROU, SRB, SVK, SVN, KOSOVO]
```
…which simply copies the value from the added file—without clarifying why the Yugoslavia-related codes are omitted.

Could you please try those two settings and see what happens?

macflo8: Applying the second suggested setting leads to the correct parametrization. So this would be a viable solution.

khaeru: Okay, good to confirm. If this works, then perhaps you can `git rebase -i main` your branch and put `d`/`drop` beside the commit that added the R12_SSP_V1.yaml file (if it was a single commit. If it was added with other files, try `e`/`edit` and adjust the commit so the file is not added).

It may also help to:
- Put a comment in R12.yaml file directly above that line, linking back to this thread or with a summary, in case others are also confused.
- Expand the header comment in R12.yaml to inform people where they could find the R12_SSP_V1.yaml from which we got that information.

macflo8:
> 
>     * Files like message_ix_models/data/material/iea_mappings/*.csv:
>       * should be stored as text, not using Git LFS.
>       * contain the same phrase "iea_mappings" in their path and filename; this could be simplified.
>       * do not appear to be mentioned anywhere in the documentation.
At the moment our biggest csv file is ~400 KB. Is it okay to store all of them as text?


>     * `message_data.tools.utilities.get_optimization_years()` does the same thing as [`ScenarioInfo.Y`](https://docs.messageix.org/projects/models/en/latest/api/util.html#message_ix_models.util.scenarioinfo.ScenarioInfo.Y). The latter is computed only once, and has tests and documentation. The former is not tested, and hits the database every time it is called. I would prefer that we take the opportunity, when moving code to `message_ix_models`, to use the better-maintained solution. Similar applies to `get_historical_years()`.

Just to be clear: This means we should refactor the functions that use these helpers and pass a `ScenarioInfo` instance to them instead, correct?

khaeru:
> At the moment our biggest csv file is ~400 KB. Is it okay to store all of them as text?

Generally yes. Git will automatically compress these files. Some further nuances:

- If the file will be manually adjusted or edited, then it is helpful to see those changes as a diff of a few lines. When stored as plain text, Git can provide this.
- If the file is wholesale output of another program and is clearly documented ("The file foobar.csv comes from running `foo bar --option=X`"), and will only ever be entirely replaced, then diffs would not be needed or very useful, and Git LFS can be used.

> Just to be clear: This means we should refactor the functions that use these helpers and pass a ScenarioInfo instance to them instead, correct?

- Only for files that are added or modified in this PR.
- It's not strictly necessary to entirely refactor or change the signature of the calling function. It's also possible as an intermediate step to replace e.g. https://github.com/iiasa/message-ix-models/blob/c8319130eab17335c4bbc8bf0eb1b9aecf92fb9d/message_ix_models/util/compat/message_data/manual_updates_ENGAGE_SSP2_v417_to_v418.py#L419 with something like:
  ```py
  model_years = ScenarioInfo(scen).Y
  ```
  This is already an improvement; it will later make it easier to see where the ScenarioInfo object can be passed/used instead of the scenario itself. Since we may want to provide tested and documented alternatives for the codes in `.util.compat.message_data`, let's not invest too much time in tinkering with them.

message_ix_models/model/material/data_methanol_new.py
macflo8: I renamed that top level function now to a imo better name and added docstrings to all functions in this module. `broadcast_years()` and `broadcast_nodes()` use `message_ix_models.util.broadcast()` to unpivot the rows that are stored in `methanol_techno_economic.xlsx`.

macflo8:
Okay thanks for confirming! I pushed the suggested changes. I think all the requested changes and comments are addressed now.

khaeru (approved):
After some offline discussion with @macflo8, this is good to merge.

In particular, the codecov/patch check failure is acceptable: per the PR description, that will be addressed via tests to be added in a subsequent PR.

PR #213, opened 2024-08-02
Transport improvements for 2024-W32

khaeru:
- [ ] **(1)** LDV base year stock ([details](https://github.com/iiasa/message-ix-models/pull/213#issuecomment-2268577218) below).
  - [x] **(1.1)** Construct total vehicle stock for 2020.
  - [x] **(1.2)** Update occupancy and VDT/vehicle/year to match (1.1) with available data.
  - [x] **(1.3)** Update shares of each technology in 2020 sales.
- [ ] **(2)** Update freight activity baseline values ([details](https://github.com/iiasa/message-ix-models/pull/213#issuecomment-2268577218) below).
- [x] **(3)** Update cost projections for t=ELC_100 & PHEV_ptrp (LDV EV) ([details](https://github.com/iiasa/message-ix-models/pull/213#issuecomment-2268577218) below)

Carried from #212:
- [x] @khaeru: Investigate `scale-1` values/pass-through of values for c=methanol.

Copied from iiasa/message_data#567:
- @r-aneeque: update freight-activity.csv for n=R12_CHN

Housekeeping:
- Update `.util.path_fallback()` and `.tools.iea.web.IEA_EWEB` to check the user's local data path for the IEA EWEB data files. This allows to use these files (by dropping in manually, or by symlinking [iiasa/message-static-data](https://github.com/iiasa/message-static-data)) without installing `message_data`.

## How to review

- Read the diff and note that the CI checks all pass.
- View [the documentation](https://iiasa-energy-program-message-ix--213.com.readthedocs.build/projects/models/en/213/transport/index.html) preview built from changes on this branch.
- **More TBA**

## PR checklist

- [ ] Continuous integration checks all ✅
- [ ] Add or expand tests; coverage checks both ✅
- [x] Add, expand, or update documentation.
- [x] Update doc/whatsnew.

PR #219, opened 2024-08-21
Fix naming of GDP and population columns in SSP data aggregation within `tools.costs`

measrainsey:
Small fix to `total_gdp` and `total_population` columns being switched in `message_ix_models.tools.costs.gdp.process_raw_ssp_data()`

In the `message_ix_models.tools.costs.gdp.process_raw_ssp_data()` function that pulls and combines SSP data in `tools.costs`, there is a block of code that converts the data in the `Computer()` into a `pandas` DataFrame, with each computer key being a column. However, the naming of the columns are switched (`total_population` and `total_gdp`). So, I just swapped them.

From what I can tell, this has zero impact on the actual costs calculations, as these DataFrames columns are not used anywhere (and are later even dropped). The calculation of GDP per capita happens before this DataFrame is created, here:

https://github.com/iiasa/message-ix-models/blob/9bb0037be15c4e76a00f12ab4c71a386ef20f6b7/message_ix_models/tools/costs/gdp.py#L87-L88

And these values seem to be correct.

I picked up on this because I am calling using the output of `message_ix_models.tools.costs.gdp.process_raw_ssp_data()` function on its own elsewhere.

## How to review

## PR checklist

- [x] Update doc/whatsnew.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -1,8 +1,11 @@
 What's new
 **********
 
-.. Next release
-.. ============
+Next release
+============
+
+- Fix and update :doc:`/api/tools-costs`
+  - Fix naming of GDP and population columns in SSP data aggregation (:pull:`219`).
 
 v2024.8.6
 =========
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -4,7 +4,7 @@ What's new
 Next release
 ============
 
-- Fix and update :doc:`/api/tools-costs`
+- Fix and update :doc:`/api/tools-costs` (:pull:`219`)
   - Fix naming of GDP and population columns in SSP data aggregation (:pull:`219`).
 
 v2024.8.6
--- a/message_ix_models/tools/costs/gdp.py
+++ b/message_ix_models/tools/costs/gdp.py
@@ -97,8 +97,8 @@ def process_raw_ssp_data(context: Context, config: Config) -> pd.DataFrame:
         return (
             pd.concat(
                 [
-                    dfs[0].to_series().rename("total_gdp"),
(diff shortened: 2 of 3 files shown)


khaeru:
I don't recall precisely, but I would guess that I put the `total_population` and `total_gdp` columns in place during work on #99 because you had other code further downstream that appeared to be using those data.

If that's since changed, and they're no longer needed, a better fix would be to remove them and the associated code entirely.

measrainsey:
> I don't recall precisely, but I would guess that I put the `total_population` and `total_gdp` columns in place during work on #99 because you had other code further downstream that appeared to be using those data.
> 
> If that's since changed, and they're no longer needed, a better fix would be to remove them and the associated code entirely.

Thanks Paul -- I am using the all of the columns in the output of `message_ix_models.tools.costs.gdp.process_raw_ssp_data()` (to make figures for writing/presenting), so for me it would be nice to keep the columns in there, even if they're not used for the module itself. But if it's causing too much clutter in the module code then we can also consider removing this part.

khaeru:
> I am using the all of the columns

Okay, then good to keep them. In that case I'd instead ask to expand the docs with half a sentence for these two columns, explaining precisely what is contained in each, so that others can maybe use them in the same way.

khaeru:
The CI failures can likely fixed by cherry-picking 9e565c9cc993dcd02f2b37c69db220c682d9def0 from #213.

khaeru (approved):
I am not sure why the `.tools.wb` cherry-picked changes are apparently not covered by the tests, since the corresponding tests do seem to run (although XFAIL).

However, LGTM and then continue with other work/clean-up, e.g. in #213 from which those changes were cherry-picked.

PR #227, opened 2024-09-10
Change decay to reach cost reduction specified by 2100 (instead of 2110) `tools.costs`

measrainsey:
Small fix to the cost decay in `tools.costs` to reach the reduction rate specified by 2100

Currently, the decay is applied so that the cost reduction rate is reached by `Config.final_year`, which is 2110 -- but this should really be 2100. So if the cost reduction rate is 10%, that 10% reduction (relative to the base year) is reached in 2110, not 2100. Therefore the 2100 value is maybe around 9% less than the base year value (or something like that).

Further down the line, the tool also takes the 2100 value and repeats it until `Config.final_year` (since 2110 isn't really a model year), so essentially the cost projections never officially reach the cost reduction rate specified (it gets very close).

This branch is derived off of the `costs/reconfigure-modules` branch, so PR #221 should be merged before this one goes through review.

## How to review

## PR checklist

- [x] Update doc/whatsnew.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -17,6 +17,7 @@ Next release
   - Replace solar and wind technologies with new ones (:pull:`206`).
   - Reorganize input files and incorporate `first_year.csv` data into `tech_map.csv` (:pull:`221`).
   - Reconfigure use and implementation of technology variants/modules to be more agnostic (:pull:`221`).
+  - Change cost decay to reach reduction percentage specified on the year 2100 (:pull:`227`).
 - Improve and extend :doc:`/material/index` (:pull:`218`).
 
   - Release of MESSAGEix-Materials 1.1.0 (:doc:`/material/v1.1.0`).
--- a/message_ix_models/tools/costs/decay.py
+++ b/message_ix_models/tools/costs/decay.py
@@ -323,7 +323,7 @@ def project_ref_region_inv_costs_using_reduction_rates(
             cost_region_2100=lambda x: x.reg_cost_base_year
             - (x.reg_cost_base_year * x.cost_reduction),
             b=lambda x: (1 - config.pre_last_year_rate) * x.cost_region_2100,
-            r=lambda x: (1 / (config.final_year - config.base_year))
+            r=lambda x: (1 / (2100 - config.base_year))
             * np.log((x.cost_region_2100 - x.b) / (x.reg_cost_base_year - x.b)),
             reference_region=config.ref_region,
         )
--- a/message_ix_models/tools/costs/decay.py
+++ b/message_ix_models/tools/costs/decay.py
@@ -278,6 +278,12 @@ def project_ref_region_inv_costs_using_reduction_rates(
     This function uses the cost reduction rates for each technology under each scenario
     to project the capital costs for each technology in the reference region.
 
+    The changing of costs is projected until the year 2100
+    (hard-coded in this function), which might not be the same
+    as :attr:`.Config.final_year` (:attr:`.Config.final_year` represents the final
+    projection year instead). 2100 is hard coded because the cost reduction values are
+    assumed to be reached by 2100.
+
     The returned data have the list of periods given by :attr:`.Config.seq_years`.
 
     Parameters


khaeru:
> (since 2110 isn't really a model year),

I don't understand this. In `year/B.yaml`, which I understand to be the default set and consistent with the current MESSAGEix-GLOBIOM global models, we have: https://github.com/iiasa/message-ix-models/blob/32cb2fd884afec2cf95878d9051cf4b33dd5af2c/message_ix_models/data/year/B.yaml#L63-L64

Does that remark mean this isn't the case?

measrainsey:
> > (since 2110 isn't really a model year),
> 
> I don't understand this. In `year/B.yaml`, which I understand to be the default set and consistent with the current MESSAGEix-GLOBIOM global models, we have:
> 
> https://github.com/iiasa/message-ix-models/blob/32cb2fd884afec2cf95878d9051cf4b33dd5af2c/message_ix_models/data/year/B.yaml#L63-L64
> 
> Does that remark mean this isn't the case?

Sorry, my wording was confusing and/or understanding of this concept is incorrect -- I meant more in the context of this tool, we're not projecting costs past 2100, but the year 2110 is still needed as an input into MESSAGEix so that's why the cost from 2100 is repeated for the 2110 time period.

measrainsey:
> I think another way of framing this is that values _are_ projected for 2110: via an assumption that values in 2110 will be identical to values in 2100. These values enter MESSAGE and will affect the model solution. That is a methodological choice that should be mentioned in the documentation.
> 
> Currently [the `Config` class has the following attributes](https://docs.messageix.org/projects/models/en/latest/api/tools-costs.html#message_ix_models.tools.costs.Config):
> 
> * Y
> * base_year
> * convergence_year
> * final_year
> * seq_years
> * y0
> 
> I am not sure whether 2100, now hard-coded in some places, is the same as any of these, or sometimes/always distinct. I also wouldn't be confident explaining the difference between these to a user. Is it possible to have a diagram showing which means what under various possible combinations of settings?

Thanks Paul, excellent points. All the year-based attributes are indeed confusing 😅 I can try to make a diagram or schematic showing the differences.

In this case, 2100 is technically none of those above (but it is part of `Config.Y` and `Config.seq_years` as those are sequences of years). Right now, 2100 is both (a) the year that the cost in the reference region needs to reach the reduction rate specified and (b) the final year of cost _change_ projection (whereas `Config.final_year` is final projection year -- 2110). 

While in our current use-case both (a) and (b) are the same, there is of course the possibility that they might be different, such as:
- user wants a technology's cost to reach 0.5 by the year 2050 
- user wants a technology's cost to reach 0.5 beyond the model horizon (such as in the year 2130)

I thought about this yesterday as I was making this PR: whether to change 2100 from being hardcoded to a parameter (or perhaps even two parameters). Not sure what would be better in this case.

khaeru:
> There is of course the possibility that they might be different, such as:
> 
> - user wants a technology's cost to reach 0.5 by the year 2050
> - user wants a technology's cost to reach 0.5 beyond the model horizon (such as in the year 2130)
> 
> I thought about this yesterday as I was making this PR: whether to change 2100 from being hardcoded to a parameter (or perhaps even two parameters). Not sure what would be better in this case.

To me, this would be easier to explain. One could say:

> Every target cost ratio value for a particular (technology, [node?], …) is associated with a target year. If not given explicitly, Config.final_year [or whichever setting] is used.
> - If this target year is before last period in the model horizon (Config.Y), then all subsequent periods will have flat/identical costs per the target.
> - If this target year is after the last period in the model horizon, the target cost ratio is not reached within the model horizon.

IOW have a *configurable default*/fallback value (e.g. 2100) while allowing the user to selectively input other values.

khaeru:
> would you recommend I add to this PR to address these concerns (and if so, in the form of additional functionality, added docs, etc)?

I think at least in the function where 2100 is newly hard-coded, the docstring should describe the function behaviour so there is no [surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) lurking for users.

Further simplification of the configuration handling is fine to defer for another PR.

measrainsey:
> I think at least in the function where 2100 is newly hard-coded, the docstring should describe the function behaviour so there is no [surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) lurking for users.
> 
> Further simplification of the configuration handling is fine to defer for another PR.

Sounds good! I've added a short blurb in the function docstrings about this -- let me know if you think additional information is needed.

https://github.com/iiasa/message-ix-models/blob/2dbdec67e094d4c70735a55c5ed977118fb991a1/message_ix_models/tools/costs/decay.py#L281-L285

glatterf42 (approved):
Looks good to me if @khaeru agrees your discussion is resolved :)

PR #228, opened 2024-09-12
Add MESSAGE model versioning names as used in AR6DB

jkikstra:
Adds model versioning names similar to those in the AR6 Scenarios Database; for documentation especially of older versions.

## How to review

Check if anything is wrong. 
Decide if we need to also add more information from those that used these model versions before merging.

### DIFF
--- a/doc/repro.rst
+++ b/doc/repro.rst
@@ -313,6 +313,12 @@ postfix
 
 Some external model names include:
 
+MESSAGE V.3
+   .. todo:: Expand with a list of cases in which this model name has been used.
+
+MESSAGEix-GLOBIOM_GEI 1.0
+   .. todo:: Expand with a list of cases in which this model name has been used.
+
 MESSAGEix-GLOBIOM 1.0
    .. todo:: Expand with a list of cases in which this model name has been used.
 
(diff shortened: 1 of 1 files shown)


doc/repro.rst
```diff
@@ -313,6 +313,12 @@ postfix
 
 Some external model names include:
 
+MESSAGE V.3
+   .. todo:: Expand with a list of cases in which this model name has been used.
```
khaeru, line 317: Given the PR title/description, it sounds like the AR6 database is a case in which this model name has been used. If so, why not record that here?

doc/repro.rst
```diff
@@ -324,6 +330,9 @@ MESSAGEix-GLOBIOM 1.1
 
    .. todo:: Expand with a list of cases in which this model name has been used.
 
+MESSAGEix-GLOBIOM 1.2
+   This version was used in <https://doi.org/10.1038/s41560-021-00904-8>, and is similar to MESSAGEix-GLOBIOM 1.1, but incorporates COVID effects and has annual timesteps from 2020 to 2025.
```
khaeru, line 334: Please add additional citations to doc/main.bib and cite; see the example of the rest of the docs.

doc/repro.rst
```diff
@@ -313,6 +313,12 @@ postfix
 
 Some external model names include:
 
+MESSAGE V.3
+   .. todo:: Expand with a list of cases in which this model name has been used.
+
+MESSAGEix-GLOBIOM_GEI 1.0
+   .. todo:: Expand with a list of cases in which this model name has been used.
```
khaeru, line 320: Ditto.

khaeru (changes requested):

PR #231, opened 2024-09-17
Fix snapshot tests

glatterf42:
The snapshot tests have been failing for a while now. Initially, if I remember correctly, there was some issue with retrieving the snapshots consistently, but that seems to be resolved. Now, one test fails because it's looking for the `PRICE_EMISSION_NEW` variable: the legacy reporting we migrated is built on top of https://github.com/iiasa/message_ix/tree/issue/723, which introduced that variable name to fix `PRICE_EMISSION`. 

While we have unfortunately still not finished the corresponding PR, the branch has progressed to a point where `PRICE_EMISSION` was actually updated instead of a new helper variable introduced. This PR thus adjusts the variable name the migrated legacy reporting expects.

## How to review

- Note that the `TEMPORARY` commit should be dropped before merging

## PR checklist

- [x] Update tests

### DIFF
--- a/message_ix_models/report/legacy/pp_utils.py
+++ b/message_ix_models/report/legacy/pp_utils.py
@@ -1865,7 +1865,9 @@ def _retr_crb_prc(ds, units):
     """
 
     # Retrieve VAR - PRICE_EMISSION
-    var = "PRICE_EMISSION_NEW"
+    # TODO iiasa/message_ix#726 is reworking PRICE_EMISSION_NEW to become the new PRICE_EMISSION
+    # Adjust the name here according to what you're running on
+    var = "PRICE_EMISSION"
     df = ds.var(var, {"type_tec": ["all"]})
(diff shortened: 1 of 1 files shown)


khaeru:
A passing thought: in theory, a similar/parallel change should have been made to the legacy reporting somewhere, perhaps on the (private) message_data branch `SSP_dev` (or whatever the name is) that is directly using the branch from iiasa/message_ix#726. Otherwise it would also break without PRICE_EMISSION_NEW.

Is there someone we can ask to point us there so we can confirm this is aligned?

glatterf42:
Good point, thanks :)
Here's the corresponding commit from two weeks ago: https://github.com/iiasa/message_data/commit/221b6de7309ec390c7d9981e5c7b554430205d5f

khaeru (approved):
Great, thanks for confirming. LGTM.

PR #236, opened 2024-09-26
Wat ssp upd1

adrivinca:
Fixes partially from the branch [water_SSP_update](https://github.com/iiasa/message-ix-models/tree/water_SSP_update) to run the nexus cooling on a R12 SSP scenario.

 a test is also added to check that the data files needed are present in the _data/water_ folder

## How to review

check that the code doe not create conflicts with other streams.
check if the style of the test is ok.
merge is there are no check errors

## PR checklist

- [ ] Update doc/whatsnew.


message_ix_models/data/water/set.yaml
```diff
@@ -370,7 +370,7 @@ nexus:
     - [loil_ppl, cooling__loil_ppl]
     - [nuc_hc, cooling__nuc_hc]
     - [nuc_lc, cooling__nuc_lc]
-    - [solar_th_ppl, cooling__solar_th_ppl]
+    # - [solar_th_ppl, cooling__solar_th_ppl]
```
glatterf42, line 373: Why did you choose to comment this out? If this is not used anymore, please delete the line. If you need it again in the future, please include a comment as to why it was commented out now and when it should be enabled again. Same below in the same file (line 911).

message_ix_models/model/water/cli.py
```diff
@@ -194,8 +194,8 @@ def nexus(context: "Context", regions, rcps, sdgs, rels, macro=False):
 
 @cli.command("cooling")
 @common_params("regions")
-@click.option("--rcps", type=click.Choice(_RCPS))
-@click.option("--rels", type=click.Choice(_REL))
+@click.option("--rcps", default="no_climate", type=click.Choice(_RCPS))
```
glatterf42, line 197:
```suggestion
@click.option("--rcps", default="no_climate", type=click.Choice(_RCPS), show_default=True, help="...")
```
Please consider including `show_default` and `help` (but please fill in a relevant explanation for `help`), which would aid people calling `mix-models cooling --help`. Same below.

message_ix_models/model/water/data/pre_processing/groundwater_harmonize.r
glatterf42, line 1: So this is what our docs say regarding "Code Style in R":
> follow the style of the existing code base.

...which is not incredibly helpful. Did you look at other R files and compare style so it is consistent?
Or, much better yet, do you happen to know tools like `ruff` for R that enforce a certain consistent style? Because if we gain more code in R, it would be great to have consistent style. 
At the moment, we even name files `.r` and `.R` at random, it seems :facepalm:

message_ix_models/tests/model/water/data/test_water_data.py
glatterf42, line 1: One line in here is marked as not covered by tests, which is odd considering this is supposed to be a test itself. However, you have not followed pytest's syntax of writing a test, so maybe nothing in here is discovered as 'a test'.
In general, pytest searches folders called "tests" (or so, you've chosen the right one), then files called "test_..." (again, right choice) and then it executes functions called `test_...` as tests, but this file doesn't contain a test function. 
So the file needs a little re-working to have a `test_` function, but you've already laid a great foundation by extracting the setup to an external function :)

message_ix_models/tests/model/water/data/test_water_data.py
```diff
@@ -0,0 +1,37 @@
+import os
+
```
glatterf42, line 2: Somewhere here, you probably want `from pathlib import Path`.

message_ix_models/tests/model/water/data/test_water_data.py
```diff
@@ -0,0 +1,37 @@
...
+from message_ix_models.util import package_data_path
+
+
+def check_files_in_folder(folder_path):
```
glatterf42, line 6:
```suggestion
def check_files_in_folder(folder_path: Path) -> None:
    """Check that every `_R11` file has a `_R12` counterpart."""
```
Always good to provide type hints, which allow automatic tools to catch errors and others to understand your code. 
Also always good to provide docstrings to help other understand your code.
Finally, you could consider renaming that function since it doesn't actually check if *general* files are in `folder_path`, but specifically compares `R11` and `R12` files. So maybe `check_R12_files_complete`?

message_ix_models/tests/model/water/data/test_water_data.py
```diff
@@ -0,0 +1,37 @@
...
+    # List of files where _R12 counterpart doesn't exist
+    missing_r12_files = []
+
+    for r11_file in r11_files:
+        # Create the equivalent _R12 file name
+        r12_file = r11_file.replace("_R11", "_R12")
+
+        # Check if the _R12 file exists
+        if r12_file not in files:
+            missing_r12_files.append(r11_file)
```
glatterf42, line 22:
```suggestion
    r12_files = {r11_file.replace("_R11", "_R12") for r11_file in r11_files}
    missing_r12_files = r12_files - files
```
Since we don't want to change the data afterwards, best to use set comprehension and subtraction here :)

message_ix_models/tests/model/water/data/test_water_data.py
```diff
@@ -0,0 +1,37 @@
+import os
+
+from message_ix_models.util import package_data_path
+
+
+def check_files_in_folder(folder_path):
+    # List all files in the folder with .csv extension
+    files = [f for f in os.listdir(folder_path) if f.endswith(".csv")]
```
glatterf42, line 8:
```suggestion
    files = {f for f in os.listdir(folder_path) if f.endswith(".csv")}
```
Since we don't want to change the data afterwards, best to use a set here.

message_ix_models/tests/model/water/data/test_water_data.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # Find files that contain "_R11" but not "6p0"
+    r11_files = [f for f in files if "_R11" in f and "6p0" not in f]
```
glatterf42, line 11:
```suggestion
    r11_files = {f for f in files if "_R11" in f and "6p0" not in f}
```
Since we don't want to change the data afterwards, best to use a set here.

message_ix_models/tests/model/water/data/test_water_data.py
```diff
@@ -0,0 +1,37 @@
...
+        if r12_file not in files:
+            missing_r12_files.append(r11_file)
+
+    # If there are missing _R12 files, print them and assert an error
+    assert not missing_r12_files, (
+        "Error: Missing _R12 files for the following _R11 files:\n"
+        + "\n".join(missing_r12_files)
+    )
+
+    print("All _R11 files have corresponding _R12 files or contain '6p0'.")
```
glatterf42, line 30:
```suggestion
    log.info("All _R11 files have corresponding _R12 files or contain '6p0'.")
```
Better than just printing such statements: use the logging functionality provided by Python and used throughout our code. You'll need to `import logging` at the top and then set `log = logging.getLogger(__name__)` also towards the top of the file to allow you to make the suggested change here.

message_ix_models/tests/model/water/data/test_water_data.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+# Define the folder path
+folder_path = package_data_path("water", "availability")
+
+# Run the file check
+check_files_in_folder(folder_path)
```
glatterf42, line 37:
```suggestion
def test_R12_files_complete():
    # Define the folder path
    folder_path = package_data_path("water", "availability")

    # Run the file check
    check_files_in_folder(folder_path)
```

glatterf42 (changes requested):
Overall, this is looking good. Aside from some questions, I would mainly like to see the new test reworked, but to have a test at all is great :)
Please let me know if my suggestions are unclear.

message_ix_models/model/water/data/pre_processing/groundwater_harmonize.r
khaeru: > Or, much better yet, do you happen to know tools like `ruff` for R that enforce a certain consistent style? Because if we gain more code in R, it would be great to have consistent style.

I am fairly sure that [`styler`](https://github.com/r-lib/styler) is that package/tool for R code, but I don't think we have ever discussed or tried to adopt it. I don't know how popular it is among our R-user colleagues.

message_ix_models/data/water/set.yaml
adrivinca: this is temporary, I did write a comment in the commit saying that I need to replace that solar_th_csp with the new technology names

glatterf42: Great! Please also include the comment here, it may be easier to find here than in the commit message that changed the line.

message_ix_models/model/water/data/pre_processing/groundwater_harmonize.r
measrainsey: Sorry for the random jump in 🙂 I use [`styler`](https://github.com/r-lib/styler) currently. I don't think it's very commonly used, but I've found it to be quite useful.

For my own personal uses, it's helpful for me to get rid of my bad lazy coding happens such as replacing all of my single quotation marks with double quotation marks.

For group uses, it's helpful  for everyone to have a consistent coding style (like Fridolin mentioned above). So, for someone like myself who doesn't like to program in `tidyverse` style, the `styler` can just help change all my code to conform to `tidyverse` style coding. 

`styler` can be installed as an addin in Rstudio (which I believe most R users use, including the ones on our team). People can easily style files one-by-one that way (such as the current file they're coding in), or there's also the option to style all .R files in a package or directory (via [`style_dir()`](https://styler.r-lib.org/reference/style_dir.html)), which would be helpful in a case like this one.

The default style used is `tidyverse` style (for example, using `<-` instead of `=` to assign objects), and if we decide to use `styler`, that would probably be the one to go with since I believe most people use that style of R coding.

For my own personal projects, I use the [`mlr`](https://github.com/mlr-org/styler.mlr) style and have modified the project's .Rprofile to use that as the default `styler`. 

I'm not sure if there is an auto-styling option, similar to what many of us use for Python code in VSCode. So for now, my process is a little more manual, by styling either a file at a time when I want or styling a directory.

khaeru: Since we use pre-commit, there is a hook for invoking styler and other R code quality tools: https://lorenzwalthert.github.io/precommit/articles/available-hooks.html#style-files

adrivinca: I did apply styler to all the scripts and I would stop here.
strangely if I change a file from .r to .R or viceversa those changes are not seen by Github and I don't think it's worth stressing too much about it :)

adrivinca:
thanks for the suggestions, I followed most of them, but I improved the test slightly differently, more in line with the other tests I hope. Now the same test can check the content in multiple folders.
I think I would continue developments in other PR and try to have this merged. thank you!

glatterf42 (approved):
Thanks, this looks great now!
From my point of view, there's only one thing left to do (provided the coverage checks now pass, which I think they should): please add this PR to `doc/whatsnew.rst` using the same style we use for the other entries there (i.e., please add a bullet point, start with an uppercase verb, describe this PR in a sentence (or more), and link the PR in the end).

adrivinca:
done, thanks

Issue #234, opened 2024-09-26
Collect pieces for SSP 2024/ScenarioMIP workflows


khaeru:
*As discussed at today's MESSAGE team meeting.*

We want to, at some point, bring together all the improvements and new code used in the SSP/ScenarioMIP workflows. This issue is to:

1. **Identify** what and where those pieces are, how they are used (in what order), etc.
2. **Decide** where they should eventually ‘land’ within `message-ix-models` or other repos.
3. **Track** as they are associated with PRs, either:
   - If atomic and separable, directly into `main`, or
   - If they need extensive clean-up that is not yet scheduled, into an intermediate, ‘collector’ branch associated with a draft PR.

## Pieces

**General/other**
- #233 by @OFR-IIASA (@macflo8 made a similar change)
- Lots of code in [iiasa/message_single_country@SSP_Dev_2023](https://github.com/iiasa/message_single_country/tree/SSP_Dev_2023)
  - **TODO** List or outline what is included.
  - @OFR-IIASA has reworked some of this code to use message_ix_models.Context objects for settings handling, which should ease future integration with message-ix-models.
- Lots of code in [iiasa/message_data@ssp_dev](https://github.com/iiasa/message_data/tree/ssp_dev)
  - **TODO** List or outline what is included.
- [iiasa/message_ix@ssp_dev](https://github.com/iiasa/message_ix/tree/ssp_dev)
  - Includes (is based on) iiasa/message_ix#726
  - Certain CPLEX settings
  - Work from @GamzeUnlu95 on MESSAGE-MACRO iteration.
    - **TODO** Confirm: is this iiasa/message_ix#430, or something else?

**Direct air capture (DAC)**
- Code in @ywpratama branch https://github.com/ywpratama/message_ix/tree/add_dac_disposal_split (this should eventually land in message-ix-models, rather than message_ix)

**Materials**
- Code on @macflo8's branch https://github.com/macflo8/message-ix-models/tree/ssp_dev-W35 

**Techno-economic parameters** (@measrainsey)
- #206
- #221
- #222
- #227

**Water/nexus**
- #223
- Another small change by @adrivinca that will be included in a small PR.

## Things to do

- Leave a comment below or edit the description to add further details.
  - Please try not to add so much that the overall list is not readable; rather add full description at the respective issue/PR, and put a ~1-sentence summary.
  - These can include future plans/intentions for next steps.
- **Open a draft PR** for any branch with associated code.
[...]

PR #242, opened 2024-10-22
Flexible setup of constraints for cooling technologies

adrivinca:
The PR changes some hard code to a flexible adjustment of constraints on cooling technologies, to improve flexibility

The cooling technologies are added as add-on to power plants (parent technologies). in some scenarios, however, the power plant are subject to constraints e.g. for phase-out, and the cooling tecnhologies `growth_activity_lo` need to be adjusted accordingly to avoid infeasibilities. Before this was done manually, now I added a function `relax_growth_constraint` that should do it automatically for each specific scenario.
I also added a test function to test the behavior of the function.

## How to review

Check the code and test are in line with the standards.
Eventually run the code on a scenario at choice and see it the model solves `mix-models --url=ixmp://ixmp_dev/%model%/%scenario% water-ix --regions=R12 cooling `, maybe @measrainsey 

## PR checklist

- [x] Update doc/whatsnew.


glatterf42:
Please make sure the `ruff` code quality check passes, then I'll review :)

That should be relatively straightforward: simply split the comments that are too long into comments on multiple lines :)

glatterf42:
If running the scenarios is truly required for a review, let's ask first before assigning people (unless you already did that in-person, then sorry, go ahead): @measrainsey, are you available to perform that review step?

measrainsey:
Hi :) @adrivinca and I have discussed this offline, so I am happy to test out the scenario run like mentioned and help provide a review.

message_ix_models/model/water/data/water_for_ppl.py
```diff
@@ -171,6 +177,91 @@ def hist_cap(x: pd.Series, context: "Context", hold_cost: pd.DataFrame) -> list:
     ]
 
 
+def relax_growth_constraint(ref_hist, scen, cooling_df, g_lo, constraint_type):
```
glatterf42, line 180: This function could really use some type hints! In particular, if you type hint `constraint_type: Literal[Union["activity", "capacity"]]` (as a [Literal](https://docs.python.org/3/library/typing.html#typing.Literal), which you might have to import from typing first (maybe `Union`, too)), you can save yourself from manually having to raise a `ValueError`: the function will automatically fail if people try to pass in invalid strings!

message_ix_models/model/water/data/water_for_ppl.py
```diff
@@ -171,6 +177,91 @@ def hist_cap(x: pd.Series, context: "Context", hold_cost: pd.DataFrame) -> list:
...
+    """
+    Function to check if the parent technologies are shut down and require
+    relaxing the growth constraint.
```
glatterf42, line 183:
```suggestion
    Checks if the parent technologies are shut down and require relaxing
    the growth constraint.
```

message_ix_models/model/water/data/water_for_ppl.py
```diff
@@ -171,6 +177,91 @@ def hist_cap(x: pd.Series, context: "Context", hold_cost: pd.DataFrame) -> list:
...
+    Parameters:
+    ref_hist (pd.DataFrame): Historical data.
+    scen (Scenario): Scenario object to retrieve parameters.
+    cooling_df (pd.DataFrame): DataFrame containing cooling technologies and
+    their parent technologies.
+    g_lo (pd.DataFrame): DataFrame containing growth constraints.
+    constraint_type (str): Type of constraint to check ("activity" or "capacity").
+
+    Returns:
+    pd.DataFrame: Updated g_lo DataFrame with relaxed growth constraints.
```
glatterf42, line 194: Thanks for adding this docstring! Please compare this to e.g. https://github.com/iiasa/message-ix-models/blob/main/message_ix_models/tools/iamc.py#L13-L26 and follow the same structure here. This will ensure that this docstring is rendered properly :)

message_ix_models/model/water/data/water_for_ppl.py
```diff
@@ -171,6 +177,91 @@ def hist_cap(x: pd.Series, context: "Context", hold_cost: pd.DataFrame) -> list:
...
+    if constraint_type == "activity":
+        year_type = "year_act"
+        bound_param = "bound_activity_up"
+    elif constraint_type == "new_capacity":
+        year_type = "year_vtg"
+        bound_param = "bound_new_capacity_up"
+    else:
+        raise ValueError(
+            "Invalid constraint_type. Must be 'activity' or 'new_capacity'."
+        )
```
glatterf42, line 205: With proper type hints above, this branch should not be necessary anymore.

message_ix_models/model/water/data/water_for_ppl.py
```diff
@@ -171,6 +177,91 @@ def hist_cap(x: pd.Series, context: "Context", hold_cost: pd.DataFrame) -> list:
...
+
+    Returns:
+    pd.DataFrame: Updated g_lo DataFrame with relaxed growth constraints.
+    """
+    if constraint_type == "activity":
+        year_type = "year_act"
+        bound_param = "bound_activity_up"
+    elif constraint_type == "new_capacity":
+        year_type = "year_vtg"
+        bound_param = "bound_new_capacity_up"
```
glatterf42, line 201:
```suggestion
    year_type = "year_act" if constraint_type == "activity" else "year_vtg"
    bound_param = "bound_activity_up" if constraint_type == "activity" else "bound_new_capacity_up" 
```
With proper type hinting above, it should be possible to shorten these conditionals. Please run `ruff`, though, and make sure the `constaint_type` is type-hinted correctly and consistently with the docstring of the function (`"capacity"` vs `"new_capacity"`).

adrivinca:
Thanks Frido and Meas.

I am not sure why this error appears, but I also had it on my machine, probably to do with some pandas update?
```
FAILED message_ix_models/tests/model/water/data/test_water_for_ppl.py::test_cool_tec[no_climate] - IndexError: index 0 is out of bounds for axis 0 with size 0
FAILED message_ix_models/tests/model/water/data/test_water_for_ppl.py::test_cool_tec[6p0] - IndexError: index 0 is out of bounds for axis 0 with size 0```

File: message_ix_models/model/water/data/water_for_ppl.py [hunk 6]
```diff
@@ -171,6 +177,91 @@ def hist_cap(x: pd.Series, context: "Context", hold_cost: pd.DataFrame) -> list:
...
+    # Look at years just after the historical year
+    bound_up1 = bound_up[bound_up[year_type] == bound_up["next_year"]]
+    # Categories that might break the growth constraints
+    bound_up1 = bound_up1[bound_up1["value"] < 0.9 * bound_up1["hist_value"]]
+    # not look ad sudden contraints after sthe starting year
+    bound_up = bound_up.sort_values(by=["node_loc", "technology", year_type])
+    # Check if value for a year is greater than the value of the next year
+    bound_up["next_value"] = bound_up.groupby(["node_loc", "technology"])[
+        "value"
+    ].shift(-1)
```
glatterf42, line 235: Just a word of caution: "off by 1"-errors are among the most prominent errors you can make while coding. So I hope you have some kind of test that ensures that shifting by this "magic number" really does exactly what you want it to do.

message_ix_models/model/water/data/water_supply.py
```diff
@@ -97,7 +97,9 @@ def map_basin_region_wat(context: "Context") -> pd.DataFrame:
         if context.type_reg == "country":
             df_sw["MSGREG"] = context.map_ISO_c[context.regions]
         else:
-            df_sw["MSGREG"] = f"{context.regions}_" + df_sw["BCU_name"].str[-3:]
+            df_sw["MSGREG"] = (
```
glatterf42, line 100: This line counts as not tested, but you're actually always setting `df_sw["MSGREG"]`, so it should count. You can make pytest recognize that by switching again to

```python
df_sw["MSGREG"] = context.map_ISO_c[context.regions] if context.type_reg == "country" else f"{context.regions}_" + df_sw["BCU_name"].str.split("|").str[-1]
```

(Can't leave suggestions on deleted lines, unfortunately. Please apply ruff here, too.)

message_ix_models/model/water/report.py
```diff
@@ -1113,7 +1113,7 @@ def report(sc: Scenario, reg: str, sdgs: bool = False) -> None:
     for ur in ["urban", "rural"]:
         # CHANGE TO URBAN AND RURAL POP
         pop_tot = sc.timeseries(variable=("Population|" + ur.capitalize()))
-        pop_tot = pop_tot[-(pop_tot.region == "GLB region (R11)")]
+        pop_tot = pop_tot[-(pop_tot.region == "GLB region (R11)")]  # ONLY R11!!!
```
glatterf42, line 1116: What does this comment mean? Please record a longer version that enables others to understand it, too :)

glatterf42 (changes requested):
All in all, good work as far as the coding goes :)
However, I'd like to see several small improvements and please check why the tests are failing on this PR and ensure they don't do that.

glatterf42:
> I am not sure why this error appears, but I also had it on my machine, probably to do with some pandas update?
> 
> ```
> FAILED message_ix_models/tests/model/water/data/test_water_for_ppl.py::test_cool_tec[no_climate] - IndexError: index 0 is out of bounds for axis 0 with size 0
> FAILED message_ix_models/tests/model/water/data/test_water_for_ppl.py::test_cool_tec[6p0] - IndexError: index 0 is out of bounds for axis 0 with size 0```
> ```

To figure this out, you can recreate a virtual environment locally that has the same pandas and numpy and Python versions installed as one of the failing workflows. Then, try to run the tests in there and use a debugger or add auxiliary output `print()` to study the variables as you go. 

What the error message says is that `years[years > x]` is a numpy array of size 0, so there's not a single element inside. Thus, when you try to access the `[0]`th element, numpy complains.

message_ix_models/model/water/data/water_for_ppl.py
adrivinca: here the shift is not on numbers, but on positions (rows) basically i am taking the 'previous' year values. And since the time steps are not constant I thought it is better then doing -5 or -10. 
However, if there are better suggestions on how to do it, welcome

message_ix_models/model/water/data/water_for_ppl.py
```diff
@@ -257,11 +348,9 @@ def cool_tech(context: "Context") -> dict[str, pd.DataFrame]:
     )
     # cooling fraction = H_cool = Hi - 1 - Hi*(h_fg)
     # where h_fg (flue gasses losses) = 0.1
-    ref_input["cooling_fraction"] = ref_input["value"] * 0.9 - 1
+    # ref_input["cooling_fraction"] = ref_input["value"] * 0.9 - 1 # probably obsolete
```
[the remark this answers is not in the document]
adrivinca: I wish that too, but I am afraid it's too much for our capacity.
anyway, I noticed that that line 396 overwrites is anyway and does the correct calculation. Maybe Muhammad forgot to remove this line?
I can also just delete it
line 396
`input_cool["cooling_fraction"] = input_cool.apply(cooling_fr, axis=1)`

message_ix_models/tests/model/water/data/test_water_for_ppl.py
```diff
@@ -201,3 +205,103 @@ def test_non_cooling_tec(request, test_context):
...
+def test_relax_growth_constraint(constraint_type):
+    # Sample data for g_lo
+    if constraint_type == "activity":
+        year_type = "year_act"
+    elif constraint_type == "new_capacity":
+        year_type = "year_vtg"
+    else:
+        raise ValueError(
+            "Invalid constraint_type. Must be 'activity' or 'new_capacity'."
+        )
```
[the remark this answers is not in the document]
adrivinca: thanks, was looking for something like this!

message_ix_models/tests/model/water/data/test_water_for_ppl.py
```diff
@@ -201,3 +205,103 @@ def test_non_cooling_tec(request, test_context):
             "year_act",
         ]
     )
+
+
+# Mock function for scen.par
```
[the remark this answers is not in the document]
adrivinca: Unfortunately, I don't have an overview on this, maybe we can bring it up at a MESSAGE meeting

message_ix_models/tests/model/water/data/test_water_for_ppl.py
```diff
@@ -201,3 +205,103 @@ def test_non_cooling_tec(request, test_context):
...
+        elif param == "bound_new_capacity_up":
+            return pd.DataFrame(
+                {
+                    "node_loc": ["R12_AFR", "R12_AFR", "R12_AFR"],
+                    "technology": ["coal_ppl", "coal_ppl", "coal_ppl"],
+                    "year_vtg": [2030, 2040, 2050],
+                    "value": [30, 15, 0],
+                }
+            )
+        return pd.DataFrame()
```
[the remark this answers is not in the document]
adrivinca: I think I need to keep the filters, because in the real case it also send filters to the function. otherwise the test breaks

glatterf42 (commented):
I'm sorry, I tried to rebase this branch on `main` to keep it up to date (since some of @measrainsey's branches were merged earlier today), but this seems to have broken the tests again.
Thanks for fixing them before! If we fix them now and add this PR to `doc/whatsnew` (and @measrainsey's review comes in), I think we're close to getting this merged :)

glatterf42:
The errors are cause by updating `sdmx1` to 2.19.0, let's see as to why exactly this happens...

khaeru:
@adrivinca given the name of this branch (`wat_SSP_upd2`) I assume this is related to #234, so I have added it in the list there. If that's not the case, please clarify.

message_ix_models/tests/model/water/data/test_water_for_ppl.py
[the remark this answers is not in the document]
adrivinca: Ideally it should be a full model, like ENGAGE scenario.
there are several parameters that are read by the water module and expectation for technologies to exist.

adrivinca:
thanks, so the `sdmx1` issue is solves, right?
@khaeru you are right, there will be many other, I will remember to add them to #234 

I am quite in a tight timeline, so I'm happy with the current changes. I might add a couple of things before merge

glatterf42 (approved):
I tried to rebase this branch on itself to reword the commit messages, but that accidentally picked up several commits not belonging here, not entirely sure why. So I'll not use that route; instead, we should `squash and merge` this PR, providing a fitting commit message.

For the future: @adrivinca, we try to have all commit messages adhere to the same form: using present tense to describe why changes are made more than what changed (which should be apparent from the code) in a short sentence. So e.g. "Fix breaking test" rather than "Fixing breaking test".

I also added this PR to `doc/whatsnew`, so I think we can merge it now, if you like. And I would appreciate not adding many things/anything of significance to a PR after it has been reviewed and approved.

adrivinca:
Thanks for helping out. just wanted to say there is still something I am adding. And realized some of the last edits are not relaxing all the constraints I wanted. So let's not merge yet, thanks

khaeru:
Understood—please notify again when you think it's ready for a re-review.

adrivinca:
Hi, I had to do some changes to my same updates.
in particular I realized that the `growth_activity_lo` constraints was causing many infeasibility issues because many parent technologies are forced to shut down in different ways (with bound activities, or lack of input data", ect). trying to accommodate all these exceptions lead to a small and not useful dataframe. That's why I released that constraint and imposed an upper bound on all cooling technologies.
in  other words, now cooling techs can be faded out quickly, but the model still cannot quickly switch to others (e.g. from once through to air cooling) withing 5% growth constraints, so that transitions should remain smooth.

SSP updated model seem to solve with unscaled infeasibilities, or normally with `scaind = 1`. But this is a know issue of the water module that we will try to tackle in another PR.

I recycled the `relax_growth_constraint` I did for `growth_activity_lo`, now for `growth_activity_up`, and adjusted the test accordingly.

One only thing that is not really good is in `set.yaml` line 958
```
    # - [solar_th_ppl, cooling__solar_th_ppl]
    - [csp_sm1_ppl, cooling__csp_sm1_ppl]
    - [csp_sm3_ppl, cooling__csp_sm3_ppl]
```
in previous scenarios (e.g. ENGAGE) the solar technology was called `solar_th_ppl`, now in the SSP updates it is different.
there is data and parameter to make both nomenclature work, but if I run, in the current case, a SSP updated scenario, the code breaks when reading the yaml file with this error
ValueError: The index set 'technology' does not have an element 'solar_th_ppl'!

on the other hand, older models might break because they do not have the new technology names. So, I am unsure on how to deal with this.

glatterf42:
Thanks for the update :)
Is `solar_th_ppl` the only name that changed from ENGAGE to the SSP Update scenarios? If not, we can look at how this was done before and do the same here.

I think the deciding question is: do you want your code to work with the old scenarios, too, or just with the SSP scenarios?
If it's supposed to work with all scenarios and nothing of the sort has been done before, then we should come up with a way to change the names we're looking for based on the scenario name. This could e.g. be a mapping of the sort:
```python
{
    "ENGAGE": "solar_th_ppl",
    "SSP Update": "some new name",
}
```
However, this looks a lot like hardcoding, which isn't great. And I can't quite believe that nothing of the sort had to be handled before, so there should be an existing solution.
If it's only supposed to work with one scenario (which could be a valid starting point even if you intend the first option eventually), it's fine to hardcode the technology name, but then your docstrings of these functions (and maybe even the top level of your module) should include warnings that they only work for the specific scenarios `[x, y, z]`.

khaeru:
I think the appropriate place for warnings is the documentation: https://github.com/iiasa/message-ix-models/blob/main/doc/water/index.rst?plain=1 , near the top.

You could add for instance:
```rst
.. note:: The current version of :mod:`.model.water` is configured to work with scenarios that have technology set elements including ``csp_sm1_ppl`` and ``csp_sm3_ppl``.
   Previous versions (in :mod:`message_ix_models` v2024.8.6 and earlier) were configured to work with scenarios including the technology ``solar_th_ppl``.
   See further discussion at :pull:`242`.
```

Some other thoughts:
- Another approach in documentation is to give a list of exact base scenario URLs with which the code is known to work. See https://docs.messageix.org/projects/models/en/latest/transport/index.html#base-scenarios
- It is indeed possible to inspect the base scenario and add the appropriate configuration based on its contents. This means that the values are not fixed in set.yaml, but rather added to the data structure after this file is read. We do that in ``.model.transport``. But I think it is likely out of scope here.
- As we can see, the presence of "ENGAGE" in a scenario name does not tell us anything useful, because very different code has been used to prepare such scenarios, and we don't have a clear record of what code was used for which (model name, scenario name, version). That's why my suggested text above says directly **what specific contents** of scenarios, rather than just alluding to "ENGAGE scenarios" which is a poorly-defined and disparate set.

PR #250, opened 2024-11-20
Confirm support for Python 3.13 and drop 3.8

khaeru:
- Check/confirm that `message-ix-models` is compatible with the new version. This includes:
  - iiasa/ixmp#544
  - iiasa/message_ix#881
  - [x] Add an explicit CI job for ixmp/message-ix 3.9.0.
  - [x] Bump GAMS version in CI from 29.1.0 to 43.4.1.
- Drop support for Python 3.8, which reached end-of-life (no more updates, even security updates) "approximately [during, before, or after] October 2024" [per PEP 569](https://peps.python.org/pep-0569/#lifespan).
   - [x] Remove any code marked as for Python 3.8 compatibility only. Depending on these changes, part or all of the functionality may remain usable with Python 3.8 if users have extreme need.
   - [x] Make any general clean-ups or improvements possible with [features available from Python 3.9](https://docs.python.org/3/whatsnew/3.9.html).
     - [x] Type hint using generics, e.g. `list[str]` instead of `from typing import List`/`List[str]`.
     - [x] Use `from collections.abc import Sequence` *et al* instead of deprecated `from typing import Sequence`.

## How to review

## PR checklist

- ~Add, expand, or update documentation.~ N/A
- [x] Update doc/whatsnew.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -4,8 +4,10 @@ What's new
 Next release
 ============
 
+- :mod:`message_ix_models` is tested and compatible with `Python 3.13 <https://www.python.org/downloads/release/python-3130/>`__ (:pull:`250`).
+- Support for Python 3.8 is dropped (:pull:`250`), as it has reached end-of-life.
 - Connect the water module to the cost module for cooling technologies (:pull:`245`).
-- Make setup of constraints for cooling technologies flexible and update solar csp tech. name (:pull:`242`). 
+- Make setup of constraints for cooling technologies flexible and update solar csp tech. name (:pull:`242`).
 - Fix the nexus/cooling function and add test for checking some input data (:pull:`236`).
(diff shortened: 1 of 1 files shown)


glatterf42 (approved):
Thanks, this is looking great :)
I hope it also explains all CI failures today, as in: GAMS was bumped here, requiring a different license, but the nightly tests and test-snapshot still wanted to use the older version requiring the older license, and failed. In that case, I'd just ask to please also adapt the [snapshot-workflow file](https://github.com/iiasa/message-ix-models/blob/main/.github/workflows/pytest-snapshots.yaml) to use the newer GAMS version :)

khaeru:
> adapt the [snapshot-workflow file](https://github.com/iiasa/message-ix-models/blob/main/.github/workflows/pytest-snapshots.yaml) to use the newer GAMS version

Ah good catch—will do.

PR #252, opened 2024-11-22
Fix snapshot test environment variables

glatterf42:
Our CI `Test snapshots` failed today, complaining that `The workflow is not valid. .github/workflows/pytest-snapshots.yaml (Line: 18, Col: 5): A mapping was not expected`.
Looking at [the docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#env), a mapping is allowed, but apparently not a nested mapping. Thus, this PR removes the `version` level in the `env` mapping (and temporarily enables `Test snapshots` to see if they run now).

## How to review

- Read the diff and note that the CI checks all pass (in particular `Test snapshot`).

## PR checklist

- [x] Update tests; coverage checks both ✅

### DIFF
--- a/.github/workflows/pytest-snapshots.yaml
+++ b/.github/workflows/pytest-snapshots.yaml
@@ -14,10 +14,9 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  version:
-    gams: "43.4.1"
-    python: "3.12"
-    upstream: main
+  gams: "43.4.1"
+  python: "3.12"
+  upstream: main
 
 jobs:
   snapshots:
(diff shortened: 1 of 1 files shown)


glatterf42:
Tests are all passing, happy to drop the TEMPORARY commit and merge this :)

khaeru (approved):
Thanks.

This is *mea culpa* via #250. Sorry for not having validated this workflow before merging that PR. Maybe we can add a pre-commit step that validates these.

khaeru:
> Maybe we can add a pre-commit step that validates these.

e.g. https://github.com/mpalmer/action-validator?tab=readme-ov-file#using-pre-commit

glatterf42:
The check is failing because of https://github.com/mpalmer/action-validator/issues/74, which confirms my suspicion: usually (see e.g. for ruff) you don't need to install the tool the hook is invoking yourself. The issue is just three weeks old, there's no associated PR yet. I'm not sure we want to do that ourselves, I don't know how long it would take to create that PR, have it merged and a new version released. It doesn't sound like a quick solution, whereas we might want to enable our snapshot tests again as soon as possible.
Maybe there's another workaround to install `action-validator` manually or maybe there's another action we could use, but we could also skip this for now.

khaeru:
> we could also skip this for now.

Yes, to be clear: my "approve" review meant "LGTM as-is", and then that suggestion was a "P.S."/tangential thought on how to avoid this in the future.

PR #218, opened 2024-08-21
Update MESSAGEix-Materials with new steel technologies 

GamzeUnlu95:
This PR updates MESSAGEix-Materials to version 1.1.0. The model changes are documented in [doc/material/v1.1.0.rst](https://github.com/iiasa/message-ix-models/blob/c9654867a0a5d7733aa6ddf2a56df1d3d9ef6b70/doc/material/v1.1.0.rst):

The PR also adds tests to the MESSAGEix-Materials build on bare RES. Solving is not possible at the moment and can be added with a follow-up PR.

## How to review

The PR is rebased on the most recent version of the main branch. The build, solve, and report functionalities are tested and work without problems. Reviewers can:

- Check the code structure
- Test the build-solve again
- Compare input gdx file with an existing model/scenario version to see if any model functionalities are missing after rebase
- Provide feedback on how to handle the versioning with the materials module

## PR checklist
- [x] Continuous integration checks all ✅
- [x] Update doc/whatsnew. ✅
- [x] Update `data/material/material.tar.gz`

PR #253, opened 2024-11-26
Convert aluminum_trade.csv from Git LFS pointer

khaeru:
This file was stored as a Git LFS pointer, i.e. the content on the branch was:
```
version https://git-lfs.github.com/spec/v1
oid sha256:b20a9561e4c08e451bf381c2733af88fa3ea0a9f776df9689b026f5434bb01d0
size 54793
```
However, the file was *not* listed in .gitattributes, at the top level or in any subdirectory. The result was that, in some cases, Git LFS would not know that it needed to 'smudge' this file —that is, fetch the LFS 'blob' and put it in the working directory.

This would in turn result in the checked-out file containing only the LFS pointer (`version https://` etc.), and code that used this file would fail. Most recently this happened in the scheduled nightly run of the "pytest" CI workflow, [here](https://github.com/iiasa/message-ix-models/actions/runs/12024191967/job/33522971771#step:10:1698).

**To resolve**, this PR:
- [x] Replaces the Git LFS pointer with the actual file contents.
- [x] @glatterf42: addressed some pandas warnings in the `.model.material` code. These warnings made it difficult to debug this failure.

## How to review

## PR checklist

- ~Add, expand, or update documentation.~ CI issue only
- [x] Update doc/whatsnew.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -23,7 +23,7 @@ Next release
   - Reconfigure use and implementation of technology variants/modules to be more agnostic (:pull:`221`).
   - Change cost decay to reach reduction percentage specified on the year 2100 (:pull:`227`).
   - Add `cooling` technology variant/module (:pull:`222`).
-- Improve and extend :doc:`/material/index` (:pull:`218`).
+- Improve and extend :doc:`/material/index` (:pull:`218`, :pull:`253`).
 
   - Release of MESSAGEix-Materials 1.1.0 (:doc:`/material/v1.1.0`).
 
--- a/message_ix_models/data/material/aluminum/aluminum_trade.csv
+++ b/message_ix_models/data/material/aluminum/aluminum_trade.csv
@@ -1,3 +1,1186 @@
-version https://git-lfs.github.com/spec/v1
-oid sha256:b20a9561e4c08e451bf381c2733af88fa3ea0a9f776df9689b026f5434bb01d0
-size 54793
+Region,Variable,Year,Value
+China,USE|Inputs|Imports,2021,0.812
+Europe,USE|Inputs|Imports,2021,4.393
+Japan,USE|Inputs|Imports,2021,1.265
+Middle East,USE|Inputs|Imports,2021,0.802
+Non Producers,USE|Inputs|Imports,2021,2.831
+North America,USE|Inputs|Imports,2021,5.248
+Oceania,USE|Inputs|Imports,2021,0.461
+Other Asia,USE|Inputs|Imports,2021,3.336
+Other Producers,USE|Inputs|Imports,2021,1.653
+South America,USE|Inputs|Imports,2021,0.5
+China,USE|Inputs|Imports,2020,0.976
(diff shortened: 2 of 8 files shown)


khaeru:
FYI @macflo8 —viewing [the file history](https://github.com/iiasa/message-ix-models/commits/main/message_ix_models/data/material/aluminum), this appeared to happen in #218.

Some tips:
- In general, we prefer to handle CSV files *not* using LFS, since they are plain text and Git compresses them fairly well anyway. As plain text, the diffs are also readable, which helps with reviews, conflict resolution, and understanding history.
- A rough heuristic is that if you can compress (Gzip, 7zip) any plain text file and it's <10 MiB, there's no need to consider Git LFS.
- In any case, files stored with Git LFS have to be marked as such in a .gitattributes file—preferably the top-level file, but also optionally a file in/scoped to a particular subdirectory.

Please reach out in the future if there's any question about these points!

glatterf42:
Thanks for this fix, @khaeru. I have now added my fixes and would still like to note two things, mainly:

- There still remain some `FutureWarnings` from pandas in material-related code (i.e. code that is run when running `tests/model/material/test_build.py`: both in line 213 and 278 of `model/material/data_util.py`, I see `FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.` I didn't want to spend too much time here digging into which of these columns or intentional, so I leave that for the future.
- While going through e.g. `model/material/util.py`, I noticed that it would likely be very worthwhile to enable `--strict` mypy here, too. Consider for example this function (originally annotated with `x: float or list[float]`, which mypy doesn't understand):
  ```python
  def exponential(x: Union[float, list[float]], b: float, m: float) -> float:
      """
      Mathematical function used in Excels GROWTH function
  
      Parameters
      ----------
      x: float or list
          domain of function
      b : float
          function parameter b
      m: float
          function parameter m
      Returns
      -------
      float
          function value for given b, m and x
      """
      return b * m**x
  ```
  I don't think `float ** list[float]` can ever work, yet there are two references to this function along the lines of `exponential([2020], ...)[0]`, which is confusing on multiple levels as we pass in a `list[int]` and cannot index the returned `float`, either. 
  I'm not sure what this implies, then. Maybe this code is untested, maybe it's never called, hopefully we don't rely on it as a crucial part of anything. Maybe every time that it is called the call is absorbed in the surrounding `try ... except RunTimeError`, but this seems like a bug that would be easily avoided by enforcing more stringent type checks :)

macflo8:
Thank you, both @khaeru and @glatterf42 for fixing issues in the materials module. Very much appreciated! The CSV LFS issue must have slipped through the cracks in #218, while I was preparing the branch for the merge. I am also tagging @GamzeUnlu95 so that both of us more are aware of this in future PRs. Sorry to mess up the nightly tests with this. I am happy to try to resolve the remaining pandas warnings in case there are some left.

macflo8 (approved):

PR #256, opened 2024-11-27
Wat ssp upd3: third update to cooling technology, with SSP differentiation

adrivinca:
This PR connects introduces a SSP configuration for cooling technologies, introduce share constraints solves some bugs like the inclusion of the correct cooling technologies for csp and geothermal plants.

## How to review

Check that code and test are in line with the standards.
Eventually run the code on a scenario at choice and see it the model solves mix-models --url=ixmp://ixmp_dev/%model%/%scenario% water-ix --regions=R12 cooling --ssp=%SSP%, maybe @measrainsey

if we can merge by Dec 2, it can be used for the ssp updates, otherwise we'll merge directly to the `ssp-dev` branch.

## PR checklist

I added these two tests:
`test_cat_tec_cooling` in `water/test_build.py`
and 
`test_cooling_shares_SSP_from_yaml` in `water/data/test_water_for_ppl.py`
- [x] Update doc/whatsnew.

[please let's not delete the branch right after merging]


glatterf42:
Already one FYI item before starting a review: the branch will automatically be deleted by GitHub after merging the PR. It will still exist on your local system, though, which is likely what you need if you want to start new work off of this branch (but keep in mind: once this branch is merged to `main`, you might as well start new work from there, which is even better than branching out too much). Should you really need the branch on GitHub, please clarify why and note that the automatic message that deletes branches also offers to restore them.

message_ix_models/model/water/build.py
```diff
@@ -3,18 +3,107 @@
...
 from message_ix_models.model.structure import get_codes
-from message_ix_models.util import package_data_path
+from message_ix_models.util import broadcast, package_data_path
 
 from .utils import read_config
 
 log = logging.getLogger(__name__)
 
 
+def cat_tec_cooling(context: Context) -> tuple[pd.DataFrame, list[str]]:
```
glatterf42, line 19: Thanks for the complete type hints :)

message_ix_models/model/water/build.py
```diff
@@ -3,18 +3,107 @@
...
+    path = package_data_path("water", "ppl_cooling_tech", FILE)
+    df = pd.read_csv(path)
+    cooling_df = df.loc[df["technology_group"] == "cooling"].copy()
+    # Separate a column for parent technologies of respective cooling
+    # techs
+    cooling_df["parent_tech"] = (
+        cooling_df["technology_name"]
+        .apply(lambda x: pd.Series(str(x).split("__")))
+        .drop(columns=1)
+    )
```
glatterf42, line 64:
```suggestion
    cooling_df = df.loc[df["technology_group"] == "cooling"].copy(deep=True)
    # Separate a column for parent technologies of respective cooling
    # techs
    cooling_df["parent_tech"] = (
        cooling_df["technology_name"]
        .apply(lambda x: pd.Series(str(x).split("__")))
        .drop(columns=1)
    )
```
I think these lines will raise a `FutureWarning` from pandas since you're setting a value on a slice of a copy of a dataframe, affecting the original dataframe as well. `deep=True` means changes on the new object won't affect the old one.

message_ix_models/model/water/build.py
```diff
@@ -3,18 +3,107 @@
...
+    # Extract unique technologies
+    sc = context.get_scenario()
+    #  get df = sc.par("input") for technollgies in cooling_df(parent_tach)
+    df = sc.par("input", filters={"technology": cooling_df["parent_tech"].unique()})
+    missing_tec = cooling_df["parent_tech"][
+        ~cooling_df["parent_tech"].isin(df["technology"])
+    ]
+    # some techs only ahve output, like csp
```
glatterf42, line 72:
```suggestion
    # some techs only have output, like csp
```

message_ix_models/model/water/build.py
```diff
@@ -3,18 +3,107 @@
...
+    ref_output: pd.DataFrame = sc.par("output", {"technology": missing_tec})
```
glatterf42, line 73: Does `sc.par()` not return a `pd.DataFrame` already? Why do we need the extra annotation?

message_ix_models/model/water/build.py
```diff
@@ -3,18 +3,107 @@
...
+        for type_tec, keywords in cooling_types.items():
+            for keyword in keywords:
+                if keyword in tech:
+                    # Add a row to the cat_tec list with type_tec and technology
+                    cat_tec_rows.append({"type_tec": type_tec, "tec": tech})
+
+    # Create the cat_tec DataFrame
+    cat_tec = pd.DataFrame(cat_tec_rows)
+
+    return cat_tec, regions_df
```
glatterf42, line 104: Kind of odd to return `regions_df`, which is a `list[str]`, but fine, I guess.

message_ix_models/model/water/build.py
```diff
@@ -30,6 +119,22 @@ def get_spec(context: Context) -> Mapping[str, ScenarioInfo]:
     remove = ScenarioInfo()
     add = ScenarioInfo()
 
+    # cooling fata included by default
```
glatterf42, line 122:
```suggestion
    # cooling data included by default
```

message_ix_models/model/water/build.py
```diff
@@ -153,21 +258,70 @@ def get_spec(context: Context) -> Mapping[str, ScenarioInfo]:
...
+            "share_cooling_air_share",
+            "share_cooling_ot_saline_share",
+        ],
+        mode="M1",
+        commodity=commodity_cool,
+        level="share",
+    ).pipe(broadcast, node_share=nodes_cooling)
+    df_share["node"] = df_share["node_share"]
+    # re order columns like this ['shares', 'node_share', 'node',
+    # 'type_tec', 'mode', 'commodity', 'level']
```
glatterf42, line 291:
```suggestion
    # re-order columns like this:
```
I'd leave out the order as it is the same as two lines below.

message_ix_models/model/water/build.py
```diff
@@ -153,21 +258,70 @@ def get_spec(context: Context) -> Mapping[str, ScenarioInfo]:
...
+    df_share = df_share[
+        ["shares", "node_share", "node", "type_tec", "mode", "commodity", "level"]
+    ]
+
+    df_list = df_share.values.tolist()
+    results["map_shares_commodity_share"] = df_list
+    # for totoal
```
glatterf42, line 298:
```suggestion
    # for total
```

message_ix_models/model/water/build.py
```diff
@@ -153,21 +258,70 @@ def get_spec(context: Context) -> Mapping[str, ScenarioInfo]:
...
+            "share_cooling_cl_fresh_tot",
+            "share_cooling_air_tot",
+            "share_cooling_ot_saline_tot",
+        ],
+        mode="M1",
+        level="share",
+    ).pipe(broadcast, node_share=nodes_cooling, commodity=commodity_cool)
+    df_share["node"] = df_share["node_share"]
+    # re order columns like this ['shares', 'node_share', 'node',
+    # 'type_tec', 'mode', 'commodity', 'level']
```
glatterf42, line 313:
```suggestion
    # re-order columns like this
```
Same as above.

message_ix_models/model/water/build.py
```diff
@@ -153,21 +258,70 @@ def get_spec(context: Context) -> Mapping[str, ScenarioInfo]:
...
+    ]
+
+    df_list = df_share.values.tolist()
+
+    results["map_shares_commodity_total"] = df_list
 
-            # Elements to add
-            add.set[set_name].extend(config.get("add", []))
+    for set_name, config in results.items():
+        # Sets  to add
```
glatterf42, line 323:
```suggestion
        # Sets to add
```

message_ix_models/model/water/data/water_for_ppl.py
```diff
@@ -31,19 +35,26 @@ def missing_tech(x: pd.Series) -> pd.Series:
...
+
+        if matched_key:
+            value = data_dic[matched_key]
+            if x["value"] < 1:
+                value = max(x["value"], value)
+            # for backwards compatibility
+            if x["level"] == "cooling":
+                return pd.Series({"value": value, "level": "dummy_supply"})
+            else:
+                return pd.Series({"value": value, "level": x["level"]})
```
glatterf42, line 54:
```suggestion
            return pd.Series({"value": value, "level": "dummy_supply"}) if x["level"] == "cooling" else pd.Series({"value": value, "level": x["level"]})
```
This avoids complexity and the need for a test case, but needs to get ruff applied to look good.

message_ix_models/model/water/data/water_for_ppl.py
```diff
@@ -85,26 +96,36 @@ def shares(
...
         cooling_fraction = hold_df[
             (hold_df["node_loc"] == col2)
             & (hold_df["technology_name"] == x["technology"])
         ]["cooling_fraction"]
-        x[col] = x[col] * cooling_fraction
 
-    results: list[Any] = []
+        # Log unmatched rows
+        if cooling_fraction.empty:
+            log.warning(
```
glatterf42, line 109: Should this be a `warning`? This looks like it's going to clutter the output potentially, so you could also think of making this `info` or `debug(ing)`.

message_ix_models/model/water/data/water_for_ppl.py
```diff
@@ -85,26 +96,36 @@ def shares(
...
+                f"No cooling_fraction found for node_loc: {col2}, technology: {x['technology']}"
+            )
+            cooling_fraction = pd.Series([0])
+
+        # Ensure the Series is not empty before accessing its first element
+        if not cooling_fraction.empty:
+            x[col] = x[col] * cooling_fraction.iloc[0]
+        else:
+            # You could choose to leave the value unchanged or apply a default
+            x[col] = x[col] * 0  # Default to 0 if cooling_fraction is empty
```
glatterf42, line 119:
```suggestion
        # You could choose to leave the value unchanged or apply a default
        # Default to 0 if cooling_fraction is empty
        x[col] = x[col] * cooling_fraction.iloc[0] if not cooling_fraction.empty else x[col] * 0  
```
This again needs ruff, but avoids tests and complexity.

message_ix_models/model/water/data/water_for_ppl.py
```diff
@@ -354,6 +450,16 @@ def cool_tech(context: "Context") -> dict[str, pd.DataFrame]:
     ref_input: pd.DataFrame = scen.par(
         "input", {"technology": cooling_df["parent_tech"]}
     )
+    # list of tec in cooling_df["parent_tech"] that are not in ref_input
+    missing_tec = cooling_df["parent_tech"][
+        ~cooling_df["parent_tech"].isin(ref_input["technology"])
+    ]
+    # some techs only ahve output, like csp
```
glatterf42, line 457:
```suggestion
    # some techs only have output, like csp
```

message_ix_models/model/water/data/water_for_ppl.py
```diff
@@ -534,6 +640,26 @@ def cool_tech(context: "Context") -> dict[str, pd.DataFrame]:
 
     results["emission_factor"] = emi
 
+    # add output for share contraints to introduce SSP assumptions
+    # also in the nexus case, the share contraints are at the macro-regions
+
+    out = make_df(
+        "output",
+        node_loc=input_cool["node_loc"],
+        # take aprt of string after "__" from icmse_df["technology_name"]
```
glatterf42, line 649:
```suggestion
        # take part of string after "__" from icmse_df["technology_name"]
```

message_ix_models/model/water/data/water_for_ppl.py
```diff
@@ -544,7 +670,7 @@ def cool_tech(context: "Context") -> dict[str, pd.DataFrame]:
     )
     df_sw["time_dest"] = df_sw["time_dest"].astype(str)
     if context.nexus_set == "nexus":
-        out = pd.DataFrame()
+        # out = pd.DataFrame()
```
glatterf42, line 673: If we don't need this line, please delete it :)

glatterf42: Otherwise, please note why it's commented and when it should be uncommented again.

message_ix_models/model/water/data/water_for_ppl.py
```diff
@@ -85,26 +96,36 @@ def shares(
...
             (hold_df["node_loc"] == col2)
             & (hold_df["technology_name"] == x["technology"])
         ]["cooling_fraction"]
-        x[col] = x[col] * cooling_fraction
 
-    results: list[Any] = []
+        # Log unmatched rows
+        if cooling_fraction.empty:
+            log.warning(
+                f"No cooling_fraction found for node_loc: {col2}, technology: {x['technology']}"
```
glatterf42, line 110: Ruff says this line is too long and it doesn't presume to split up long string correctly, so please make this a two-line string manually to pass the code quality check.

glatterf42 (changes requested):
Thanks for this PR :)

I've left several notes and comments in-line. I know this is not our current practice, so I don't mean to blame you, but please be aware that the ideal workflow would be slightly different: you open a PR and work on it until all tests pass and only request a review once that's the case. In between those, you could even think of reviewing your code yourself to e.g. catch typos or add explanatory comments to changes that you think might warrant them.

adrivinca:
yes sorry for the miscommunication. The PR is not ready for review yet, I should have written it

message_ix_models/data/water/ssp.yaml
```diff
@@ -0,0 +1,61 @@
...
+  - CPA
+  - CHN
+  - RCPA
+  - MEA
+  - PAS
+scenarios:
+  SSP1:
+    cooling_tech:
+      globalNorth:
+      # constraints on for the globalNorth
```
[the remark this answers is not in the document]
adrivinca: for, mistake

message_ix_models/data/water/technology.yaml
```diff
@@ -50,8 +42,8 @@ nexus:
 
   bio_istig__ot_fresh:
     description: >-
-      # TODO fill in description of bio_istig cooling by once through cooling
-      # technology using freshwater
+    # TODO fill in description of bio_istig cooling by once through cooling
```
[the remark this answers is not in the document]
adrivinca: I did no changes here, just moved things aroud. I actually think some of these descriptions are slightly incorrect (like the input declaration) and I noted it down to go through it, but no time these week(s).

message_ix_models/model/water/build.py
adrivinca: I am not sure, but also in `water_for_ppl` other parameters were introduced like this, so I just did the same

message_ix_models/model/water/data/water_for_ppl.py
adrivinca: changed to info. thanks!

message_ix_models/data/water/technology.yaml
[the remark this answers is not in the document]
glatterf42: There should be enough time to make it `once-through`, though, right?

message_ix_models/model/water/data/water_for_ppl.py
glatterf42: Complexity as measured by code quality tools, yes. Writing is as I suggested enables the code to understand that all you're doing is returning a `pd.Series`, even though the contents differ. Writing it as you did means the tests (and potential readers) will have to check two arms of in if branch to see what you return.

adrivinca:
thanks a lot for the inputs, now I think I brought the PR in a good state, adding tests and test should pass.

message_ix_models/data/water/ppl_cooling_tech/cool_techs_country_share.csv
glatterf42, line 1: For this and some other `.csv` files, I think when I first viewed this PR, they were uploaded normally, but they now seem to be git lfs files. This is not how we want to treat csv files, see [this recent comment](https://github.com/iiasa/message-ix-models/pull/253#issuecomment-2499989811) by @khaeru. Could you please revert that?

glatterf42 (commented):
Thanks for the updates and new tests, this is beginning to look good :) 
Unfortunately, the tests are only passing with version 3.9.0 of message_ix or newer, the reason being that [`make_df()` could only be used for `parameters` before that, not `sets`](https://docs.messageix.org/en/stable/whatsnew.html#all-changes). This leaves you two options on how to resolve this: you could either build your own functionality for this or you can mark your function(s) (that rely on `make_df()` as requiring a minimum version of 3.9.0 for message_ix.

For rebuilding functionality, you could with an `if` check so that code that is new enough uses `make_df`, but older code doesn't. For this older code, you could try to manually duplicate the [steps performed by `make_df`](https://github.com/iiasa/message_ix/blob/c4fde4465713bb6ebd496aafc0ea2b26b495b010/message_ix/util/__init__.py#L108-L146), but you could also just construct your own `pd.DataFrame` in another way.
For marking your code as requiring a minimum version, we have [this decorator function](https://github.com/iiasa/message-ix-models/blob/464e579d2f9214d9866dd39fb167d1fe8908b297/message_ix_models/util/__init__.py#L535) already and [this is how you would use it](https://github.com/iiasa/message-ix-models/blob/464e579d2f9214d9866dd39fb167d1fe8908b297/message_ix_models/model/snapshot.py#L104). Please keep in mind to apply this to every function calling the `make_df` function directly or another function that calls it indirectly, including tests.

Of course, the first approach has the advantage that your code is more compatible with other code, while the second approach may be faster to implement right now.

message_ix_models/data/water/ppl_cooling_tech/cool_techs_country_share.csv
khaeru: …and IMO this is the easiest and fastest thing to do, and probably not a problem unless you consciously want to support certain users who will have message_ix ≤ 3.8 and not be able to upgrade.

adrivinca: Thanks a lot, I did update for `make_df` and wanted to clean up the data files.
I noticed that a file in particular might be large (4mp) and it is not used anymore 
`\message_ix_models\data\water\ppl_cooling_tech/POWER_PLANTS_2016_Raptis.csv`
maybe I remove this completely and just leave the others as normal csv?

khaeru: Yes, please go ahead to remove any data that's no longer in use. As long as it showed up on `main` at any point, it is never lost, and we can always go back and 'resurrect' it from the Git history as needed. In the meanwhile keeping the repo and package smaller helps things be lightweight and performant for everybody.

And non-LFS for the other CSV files: yes please.

glatterf42:
Thanks for these fixes! As far as I can tell, the tests are now failing because the `csv` files stored under https://github.com/iiasa/message-ix-models/tree/wat_SSP_upd3/message_ix_models/data/water/delineation are also stored as Git LFS files. They are read in as usual `csv` files, but pandas just reads in the string specifying how Git LFS would load the contents of the file, not the actual contents. In light of the other changes made here, I think the best solution would be making sure that all `/data/water/*.csv` files are stored as plain `csv` files.

adrivinca:
Hi, I noticed that all test solved after I replaced `mafe_df()` but then fail again in mac and Ubuntu OS after converting the LFS to normal csv files.
It seems all test pass on Windows, so could anybody help me understand the problem on other OS please?

Other than that I rebased, updated doc and Whatsnew, So I would be done

doc/whatsnew.rst
```diff
@@ -4,6 +4,7 @@ What's new
 Next release
 ============
 
+- Introduce a SSP configuration for cooling technologies with share constraints (:pull:`256`).
```
glatterf42, line 7:
```suggestion
- Introduce an SSP configuration for cooling technologies with share constraints (:pull:`256`).
```
Isn't it `an`? I think `a`/`an` are based on sound and if you say `SSP` (the abbreviation), you say `es es pee`, sort of.

glatterf42 (approved):
Thanks, this looks good to me now, provided tests are passing :)

adrivinca:
Yes, thanks for the support @glatterf42 and @khaeru !
next steps (probably next year) will be re-organizing the config options for the function, with Paul's help and then moving to the nexus function.

khaeru:
Please go ahead and merge if you're satisfied @glatterf42 —I don't have bandwidth at the moment to review. If #235 needs manual intervention to rebase on `main` after this goes in, I will help there.

glatterf42:
Well, no objections from my side, but what about this part of the review:

> Eventually run the code on a scenario at choice and see it the model solves mix-models --url=ixmp://ixmp_dev/%model%/%scenario% water-ix --regions=R12 cooling --ssp=%SSP%, maybe @measrainsey

Is this step required to be performed by someone else? If we're just verifying it solves without errors, I expect you have confirmed this already, @adrivinca? If so, we can merge, otherwise, who should confirm it?

adrivinca:
yes, let's not to use too many people time, I have checked it myself. All good.
thanks

PR #261, opened 2024-12-04
Change in LED setup after group feedback

adrivinca:
minimal change to the SSP setup upon Volker's suggestion


khaeru:
Incorporated in #235 by cherry-pick.

adrivinca:
I also had to add a last minute change in the data. I missed some csp technologies (having 16 technologies with irregular names does not help automatize the parametrization ..)

I would still like to merge this in main and not just cherry-pick in ssp-dev because I don't want to have divergent branches like in message_data, at least for the water implementation.

PR #262, opened 2024-12-04
Change in SSP parameters and add missing csp data

adrivinca:
same as #261 but it seems I cannot reopen that Pr because I rebased the branch
here we have:

1. minimal change to the SSP setup upon Volker's suggestion.
2. I also had to do a last minute change in the data. I missed some csp technologies (having 16 technologies with irregular names does not help automatize the parametrization ..)

I would still like to merge this in main and not just cherry-pick in ssp-dev because I don't want to have divergent branches like in message_data, at least for the water implementation.

anyway, not tests added, no changes in doc, but let me know if I should do a whatsnew

thanks


khaeru (approved):
> let let me know if I should do a whatsnew

Did this while reviewing. LGTM! I'll merge once checks pass.

PR #268, opened 2024-12-10
Wat ssp upd4

adrivinca:
**Required:** write a single sentence that describes the changes made by this PR.

Fixes for the SSP update.
we noticed nuc_lc cooling technology was not correctly parametrized. I did not notice it because in all the scenarios I tested the technology was not active, but in some scenario it is, before 2020

Oliver suggested to pre-filter the `remove.set` as currently it takes lot of time just to check if the set is there

## How to review

Still need to check if it correctly working on some scenarios with 2020

## PR checklist

- [ ] whatsnew.


message_ix_models/model/water/build.py
glatterf42, line 1: So this is a `for` inside a `for` inside a `for`? I'm surprised that this is the faster solution, it doesn't seem very efficient. Without spending more time, I'm not sure how to improve this, though, and if you say it's an improvement, then it's fine with me :)

glatterf42 (approved):
Looks good to me, thanks :)

message_ix_models/model/water/build.py
adrivinca: the thing is that for any item to be `message_ix_models.model.build.apply_spec()` remove function checks if elements exist and proceeds or says that there are no elements. This is very slow.
now we just exclude from the remove list in advance sets that we know are not in the scenario, so that they do not get sent to `apply_spec()`

glatterf42: If this is a serious problem and/or you know a solution for it, please record it in a new issue. This will raise awareness and eventually lead to a solution :)

khaeru (changes requested):
@glatterf42 @adrivinca —I see there was previously an approving review here, but:
- I don't understand what the title means.
- The tests appear to be failing.
- The PR checklist is missing items and is not complete.
- The branch needs to be rebased.

So I am just leaving this review to request/record that these things need to be addressed so we can merge this PR.

adrivinca:
Hi, there has been more things I added after and more to come. i can update the description and notify where everything is ready for review.
thanks

adrivinca:
@glatterf42 I think now the PR is ready to be reviewed.
let's see if texts work, I removed one because I got rid of the tested function, but added a new one.

@khaeru I see there is a requested change from you, but maybe it's obsolete? I cannot see the content

doc/whatsnew.rst
```diff
@@ -126,6 +126,7 @@ Water/Nexus
 - Connect :mod:`.model.water` to :mod:`.tools.costs` for cooling technologies (:pull:`245`).
 - Introduce an SSP configuration for cooling technologies with share constraints (:pull:`256`).
 - Change SSP parameters and add missing CSP data (:pull:`262`).
+- Additional final constraints and data on cooling technologies for the SSP runs (:pull:`268`).
```
glatterf42, line 129:
```suggestion
- Add additional final constraints and data on cooling technologies for the SSP runs (:pull:`268`).
```

All these lines should start with verbs to announce what they _do_.

message_ix_models/model/water/build.py
```diff
@@ -103,6 +103,152 @@ def cat_tec_cooling(context: Context) -> tuple[pd.DataFrame, list[str]]:
...
+    regions_df : list
+        List of region nodes
+    commodity_mapping : dict, optional
+        If provided, maps each share_key to a specific commodity.
+
+    Returns:
+    -------
+    list
+        List of share mapping rows
+    """
```
glatterf42, line 125: I would love if we could get used to type hinting new code that we write :)
This serves both as documentation and clarification of intent. And furthermore, it makes rewriting and maintaining the code that much easier! Every time one tries to use it in a way it's not supposed to be used, mypy immediately stops one :)

In this case, you already have spelled out what these parameters should be in the docstring, so please include this in the function definition, too:

```python
def share_map_cool(share_keys: list[str], type_tec_keys: list[str], regions_df: list[str], commodity_mapping: dict[str, str]) -> pd.DataFrame:
```

For this annotation, I guessed the types of the `list[]` and `dict[]` based on your description. You'll note that `regions_df` is a misnomer if it truly is a `list[str]`; it should probably only be called `*_df` if it is a `pd.DataFrame` and just `regions` if it's just a `list`. 
Similarly, I'm reasonably sure the return value is a `pd.DataFrame`, not a `list`, so the docstring needs adapting if that is correct :)

message_ix_models/model/water/build.py
```diff
@@ -103,6 +103,152 @@ def cat_tec_cooling(context: Context) -> tuple[pd.DataFrame, list[str]]:
...
+    df_share = pd.DataFrame(expanded_rows).pipe(broadcast, node_share=regions_df)
+    df_share["node"] = df_share["node_share"]
+    return df_share[
+        ["shares", "node_share", "node", "type_tec", "mode", "commodity", "level"]
+    ]
+
+
+def cat_tec_cooling_calib(
+    context: Context,
+) -> tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame]:
```
glatterf42, line 161: These type hints are great, for example :)

message_ix_models/model/water/data/pre_processing/calculate_ppl_cooling_technology_shares.R
glatterf42, line 1: Please note: I'm not familiar with R, so can't really review this file.

message_ix_models/model/water/report.py
```diff
@@ -4,11 +4,15 @@
...
 import pandas as pd
 import pyam
-from message_ix import Reporter, Scenario
 
-from message_ix_models.report.legacy.iamc_report_hackathon import (
+# from message_ix_models.report.legacy.iamc_report_hackathon import (
+#     report as legacy_report,
+# )
+from message_data.tools.post_processing.iamc_report_hackathon import (
     report as legacy_report,
```
glatterf42, line 12:
```suggestion
from message_data.tools.post_processing.iamc_report_hackathon import (
    report as legacy_report,
```

message_ix_models/model/water/report.py
```diff
@@ -105,12 +110,14 @@ def collapse_callback(df):
...
+
+        rep_dm.require_compat("pyam")
+        key = rep_dm.add("demand", "as_pyam", rename=rename, collapse=collapse_callback)
 
-        key = rep_dm.convert_pyam("demand", rename=rename, collapse=collapse_callback)
         # Making a dataframe for demands
-        df_dmd = rep.get(key).as_pandas()
-        df_dmd = df_dmd.drop(columns=["exclude"])
+        df_dmd = rep_dm.get(key).as_pandas()
+        # df_dmd = df_dmd.drop(columns=["exclude"])
```
glatterf42, line 120: If you're keeping commented lines, it's always good to explain why and what needs to happen to uncomment them again.

message_ix_models/model/water/report.py
```diff
@@ -271,6 +278,152 @@ def multiply_electricity_output_of_hydro(
     return report_iam
 
 
+def pop_water_access(sc: Scenario, reg: str, sdgs: bool = False) -> None:
```
glatterf42, line 281:
```suggestion
def pop_water_access(sc: Scenario, reg: str, sdgs: bool = False) -> pd.DataFrame:
```

Not sure how mypy accepted this, but the function clearly returns a `pd.DataFrame` and not `None`. Is mypy enabled for the water module?

message_ix_models/model/water/report.py
```diff
@@ -1295,7 +1349,7 @@ def report(sc: Scenario, reg: str, sdgs: bool = False) -> None:
     report_iam.filter(variable=varsexclude, unit="unknown", keep=False, inplace=True)
     # prepare data for loading timeserie
     report_pd = report_iam.as_pandas()
-    report_pd = report_pd.drop(columns=["exclude"])
+    # report_pd = report_pd.drop(columns=["exclude"])
```
glatterf42, line 1352: Same as above: please explain why this commented line is needed.

message_ix_models/model/water/report.py
```diff
@@ -1326,7 +1381,7 @@ def report(sc: Scenario, reg: str, sdgs: bool = False) -> None:
     df_unit = df_unit[~df_unit["variable"].str.contains("Investment")]
     df_unit_inv = df_unit_inv.as_pandas()
     report_pd = pd.concat([df_unit, df_unit_inv])
-    report_pd = report_pd.drop(columns=["exclude"])
+    # report_pd = report_pd.drop(columns=["exclude"])
```
glatterf42, line 1384: Same as above: please explain why this commented line is needed.

message_ix_models/tests/model/water/data/test_water_for_ppl.py
```diff
@@ -213,93 +213,51 @@ def test_non_cooling_tec(request, test_context):
...
-            "year_act": [2030, 2040, 2050, 2030],
-            "time": ["year", "year", "year", "year"],
-            "value": [-0.05, -0.05, -0.05, -0.05],
-            "unit": ["%", "%", "%", "%"],
-        }
-    )
-
-    # Sample data for ref_hist
-    ref_hist = pd.DataFrame(
+def test_apply_act_cap_multiplier(param_name, cap_fact_parent, expected_values):
```
glatterf42, line 237: Did you know? One can even type hint tests and it will make them safer, too :)
Based on your parametrization above, this should be

```suggestion
def test_apply_act_cap_multiplier(param_name: str, cap_fact_parent: Optional[pd.DataFrame], expected_values: list[float]) -> None:
```

(where you might have to import `Optional` from typing).

glatterf42 (commented):
Looks good in general, thanks :)
A few comments in-line and I'll check the tests next :)

glatterf42:
Did you check the error message produced by pytest? E.g. on ubuntu-latest-py3.13-upstream-main (which I consider the most important check), I see this:

```python
 _____ ERROR collecting message_ix_models/tests/model/water/test_report.py ______
ImportError while importing test module '/home/runner/work/message-ix-models/message-ix-models/message_ix_models/tests/model/water/test_report.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../../.local/share/uv/python/cpython-3.13.2-linux-x86_64-gnu/lib/python3.13/importlib/__init__.py:88: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
message_ix_models/tests/model/water/test_report.py:8: in <module>
    from message_ix_models.model.water.report import report_full
message_ix_models/model/water/report.py:11: in <module>
    from message_data.tools.post_processing.iamc_report_hackathon import (
E   ModuleNotFoundError: No module named 'message_data'
```

So the error happens while pytest tried to import `model/water/report.py`, which it does before trying to run the actual tests. And the error message seems pretty clear: `message_data` is not installed and it shouldn't be: message-ix-models is open source, while message_data is closed source, so message-ix-models should not depend on message_data if we want to keep it that way. That is why we moved _one version_ of the reporting in message_data to [message-ix-models](https://github.com/iiasa/message-ix-models/blob/main/message_ix_models/report/legacy/iamc_report_hackathon.py). If this version of the function doesn't suit you, feel free to adapt it or rewrite it's behaviour using the new and improved genno-based system, but don't import it from message_data.

message_ix_models/model/water/report.py
glatterf42: I missed that just looking at the code: see comment below, but this import shouldn't be in there at all.

message_ix_models/model/water/report.py
khaeru: It is not: https://github.com/iiasa/message-ix-models/blob/a45fce49fe3aa17ddb9d83f8e5b550accc0090b4/pyproject.toml#L113-L117

glatterf42: Thanks for checking, that's a shame :neutral_face:

khaeru: With a new colleague (Vignesh) coming onboard to work on nexus/water topics, perhaps some of the added bandwidth could be reserved for improving code quality through that and other TODOs.

adrivinca:
Thanks, concerning the reporting. the module does not require a specific version of the legacy. This is just based on what scenario is the nexus module attached to. to report the water part the new reporting is used.
but I am not aware of the compatibility of the message_ix_models legacy_report with the most recent MESSAGE scenarios. Therefore I would keep the connection to message_data, using one of the suggestions from Paul.

message_ix_models/model/water/report.py
adrivinca: I think I had to comment because of some updated in pandas that made the previous formulation obsolete. But I am not 100% sure, that 's why I kept the old version for safety

message_ix_models/model/water/report.py
adrivinca: thanks, but in theory, if I have the extension on VSC should it do my typewriting itself?
I have mypy type checker (the open source mypy breaks..)

message_ix_models/model/water/data/pre_processing/calculate_ppl_cooling_technology_shares.R
adrivinca: ja, I think we are not testing the R script for now. and slowly we'll shift to python completely

message_ix_models/model/water/report.py
glatterf42: I'm also using the mypy type checker extension :)
But I don't see anywhere in their docs that they're automatically adding types to something. I do also have the Pylance extension and this infers some types even for untyped functions in the signature help it displays when hovering over a function name (for example). However, this does not mean that these functions are type checked! Pylance is just guessing the types and nothing makes sure these guesses are enforced. So if you want your functions to have type hints (for all the many good reasons), you have to add them manually, I think. 
But as I mentioned before: you already spell out which types the parameters should have in the docstring, so it's not a lot of work :)

message_ix_models/model/water/build.py
```diff
@@ -103,6 +103,159 @@ def cat_tec_cooling(context: Context) -> tuple[pd.DataFrame, list[str]]:
     return cat_tec, regions_df
 
 
+def share_map_cool(
+    share_keys: list[str],
+    type_tec_keys: list[str],
+    regions_df: list[str],
+    commodity_mapping: dict[str, str] = None,
```
glatterf42, line 110:
```suggestion
    commodity_mapping: Optional[dict[str, str]] = None,
```

This would be the correct type hint if the default is `None`. Otherwise, mypy would complain because `None` is not compatible with `dict[str, str]`.

adrivinca:
thanks for the quick feedback and review! now test pass. apart from the codecov check.
and yes, we will work with Vignesh to improve the code in several parts, hopefully in the reporting too. But probably we will start from other sections (not cooling technologies) that are more urgent for other projects.

Please let me know if this can be merged today for @OFR-IIASA to be able to run simply rebasing `ssp-dev`.
Thank you!!

glatterf42:
I've rebased the branch to rewrite some commit messages. Please note for the future that we want all commits to start with a capitalized verb, e.g. "Fix xyz", not "Fixing xyz". 
Next, I looked at the missing lines to see if we could easily cover them with tests. Unfortunately, that does not seem to be the case since we still don't have what you promised in https://github.com/iiasa/message-ix-models/pull/106: a test scenario set up in such a way to run the water code on it. So from my view, we could merge the PR without tests again, but I don't want this to become a theme: opening a PR, working on it for some time, and then postponing the tests because the PR should ideally be merged ASAP. So I'm inclined to say: for this one, it's okay one last time, but the next water PR should add tests as promised about two years ago (https://github.com/iiasa/message-ix-models/pull/88 and https://github.com/iiasa/message-ix-models/pull/93). Please chime in @khaeru (and adjust your requested changes if you think we can now merge this PR).

khaeru:
This PR seems to have triggered [this `ruff` error](https://github.com/iiasa/message-ix-models/actions/runs/14052087870/job/39344232016#step:5:43):
```python
message_ix_models/model/water/data/water_for_ppl.py:283:5: C901 `cool_tech` is too complex (12 > 11)
    |
281 | # water & electricity for cooling technologies
282 | @minimum_version("message_ix 3.7")
283 | def cool_tech(context: "Context") -> dict[str, pd.DataFrame]:
    |     ^^^^^^^^^ C901
284 |     """Process cooling technology data for a scenario instance.
285 |     The input values of parent technologies are read in from a scenario instance and
```

I am not sure why this wasn't picked up by the same code quality CI job on the branch here, which passed. So two actions, for which we can have separate issues/PRs:
- Reduce complexity of `cool_tech()` (preferred) or add an exemption inline or in pyproject.toml.
- Investigate the way the workflow functions and see if there is anything to be fixed.

glatterf42:
As for investigating the workflow: The `action/checkout` step shows that `a45fce49fe3aa17ddb9d83f8e5b550accc0090b4` was checked out, which is [a commit on main](https://github.com/iiasa/message-ix-models/commit/a45fce49fe3aa17ddb9d83f8e5b550accc0090b4). So maybe we just didn't adjust the code quality check and this doesn't use a branch's version of the code, but the one on main?

khaeru:
We identified that there's a typo in the YAML file, note `outputs` vs `output`:
https://github.com/iiasa/message-ix-models/blob/a0d542014d8a6908410137ec6662202c7bf8cf66/.github/workflows/pytest.yaml#L111
https://github.com/iiasa/message-ix-models/blob/a0d542014d8a6908410137ec6662202c7bf8cf66/.github/workflows/pytest.yaml#L192

Fixing the latter to match the former should hopefully resolve the issue.

adrivinca:
hi, I am a bit lost. is it a fix needed in the water module or in pytest?
or will it imply fixes in both?
thanks

PR #271, opened 2024-12-18
Move DAC module's `tools.costs` assumptions into their own CSV files

measrainsey:
PR to reorganize @ywpratama's DAC costs inputs 

When @ywpratama initially added the DAC costs assumptions for #235, the new DAC technologies were added to the energy module's cost input CSVs, as at the time the costs tool didn't have the ability to take custom cost assumptions for the modules. Since then, #255 was merged into main and now the costs tool allows for module-specific costs assumptions. I have moved the DAC technologies' reduction rate categories and scenario reduction assumptions out of the energy module's directory into the DAC module's directory.

Additionally, this PR should help to address the issue raised by @khaeru in his comment [here](https://github.com/iiasa/message-ix-models/pull/235#issuecomment-2546330758): due to another cost reduction category being added for other technologies but not the DAC technologies, this led to a problem where the DAC technologies were missing a value in the input CSV.

I also added the `dac` module to the type hints in functions that require module as an input.

## How to review

For @ywpratama : Can you confirm if these cost assumptions are correct for the DAC module and nothing was incorrectly transferred over?

For @khaeru and/or @glatterf42 : In case you want to take a look as well?

## PR checklist

None I believe as this is won't be merged to `main`.


glatterf42 (approved):
Looks good from my side :)

ywpratama:
Also looks good from my side. Thanks, Meas!

PR #279, opened 2025-01-16
Add documentation for running MESSAGEix on UniCC

measrainsey:
WIP: migrate my internal guide and documentation on how to run MESSAGEix on the cluster (UniCC) and make some modifications

## How to review

## PR checklist

- [x] ~Add or expand tests; coverage checks both ✅~
- [x] Update doc/whatsnew.


doc/distrib/unicc.rst
```diff
@@ -0,0 +1,951 @@
...
+can be checked:
+
+.. code:: bash
+
+   → cat gamslice.txt 
+   GAMS_Demo,_for_EULA_and_demo_limitations_see___G240131/0001CB-GEN
+   https://www.gams.com/latest/docs/UG%5FLicense.html_______________
+   1496554900_______________________________________________________
+   0801346905_______________________________________________________
+   DC0000_______g_1_______________________________C_Eval____________
```
glatterf42, line 503: Even if it's a demo, I'm not sure we should post the contents to our docs.

doc/distrib/unicc.rst
```diff
@@ -0,0 +1,951 @@
...
+
+.. code:: bash
+
+    #SBATCH --time=20:00:00
+    #SBATCH --mem=100G
+    #SBATCH --nodes=1
+    #SBATCH --ntasks=1
+    #SBATCH --cpus-per-task=16
+
+I usually request lots of run time (20 hours) and lots of memory (100 GB) because I don't want my job to for those reasons.
```
glatterf42, line 934:
```suggestion
I usually request lots of run time (20 hours) and lots of memory (100 GB) because I don't want my job to fail for those reasons.
```

doc/distrib/unicc.rst
```diff
@@ -0,0 +1,951 @@
...
+
+However, this will show a lot of information, so it might be better to
+run a more specific command like:
+
+.. code:: bash
+
+   sacct --format=jobid,MaxRSS,MaxVMSize,start,end,CPUTimeRAW,NodeList
+
+
+Resources to request for reducing MESSAGEix run time
```
glatterf42, line 914: It's great to have a section like this, but if the cluster becomes more heavily used, there are more considerations for setting these parameters: in general, slurm will try to optimize its load so that it's running almost always at full capacity. This makes it harder to request lots of resources for extended amounts of time because the system will not be able to squeeze in such a task with other tasks, thus it will likely take longer for the task to be picked up. 
I don't know if UniCC is already experiencing the kind of load to make these considerations necessary, but if/once it does, we would likely want to optimize these parameters further to request the minimum necessary that still reduces the run time :)

glatterf42 (commented):
Thanks for this extensive guide! Skimming over it, I thought of just a few notes :)

doc/distrib/unicc.rst
khaeru: I've removed this via rewriting the commit.

khaeru (approved):
Thanks @measrainsey for this valuable contribution!

Some items for follow up:
- One thing I believe we discussed orally was to organize this and the other HOWTO guide in their own docs section.
- I would also like to make sure that the guide teaches users where to find information in the UnICC docs and by using e.g. the man pages for `sbatch` and similar commands.

But LGTM and make those improvements while addressing #284 et al.

PR #281, opened 2025-01-21
Update CI and code quality

khaeru:
- Use astral-sh/setup-uv in "pytest" CI workflow, parallel to iiasa/ixmp#555 and iiasa/message_ix#911.
- Update ruff version and format some files.
- Update mypy configuration as suggested at python/mypy#13916, i.e. invoke mypy on all files, even those not modified in a particular commit.
  - Address many mypy errors in common code and in `.model.transport`.
  - Mark other code via pyproject.toml as excluded, for now.

## How to review

@glatterf42:
- Read the diff for non-code files. (No need to review all the typing changes.)

All others: please ignore! Changes to type hints only.

## PR checklist

- ~Add, expand, or update documentation.~ N/A, CI changes only
- ~Update doc/whatsnew.~


.github/workflows/pytest.yaml
```diff
@@ -1,19 +1,44 @@
 name: Test
 
 on:
+  # Uncomment this entry only to debug the workflow
```
glatterf42, line 4: None of the following entries seems commented, is this on purpose?

message_ix_models/tools/exo_data.py
```diff
@@ -43,9 +44,12 @@ class ExoDataSource(ABC):
     key: Optional[Key] = None
 
     #: Name for the returned :class:`.Key`/:class:`.Quantity`. Optional. See
-    # :meth:`get_keys`.
+    #: :meth:`get_keys`.
     name: str = ""
 
+    #: Primary measure.
+    measure: str = ""
```
glatterf42, line 51:
```suggestion
    measure = ""
```
Shouldn't this be enough for any type checker to infer the type of `measure`?

.github/workflows/pytest.yaml
```diff
@@ -1,19 +1,44 @@
...
+  python-version: "3.13"
+
 jobs:
+  check:
+    name: check permissions
+    runs-on: ubuntu-latest
+    steps:
+    - if: >
+        github.repository != github.event.pull_request.head.repo.full_name &&
+        ! contains(github.event.pull_request.labels.*.name, env.label)
```
glatterf42, line 34: Doesn't this also need 

```
!(
          github.event_name == "schedule"
          || github.repository == github.event.pull_request.head.repo.full_name
          || contains(github.event.pull_request.labels.*.name, env.label)
        )
```

like the message workflow?

glatterf42 (approved):
Thanks, looks good to me. I love that we're type hinting more stuff :)

message_ix_models/tools/exo_data.py
khaeru: Hm, true. Thanks!

PR #283, opened 2025-01-23
Fix envvar name in snapshot CI

glatterf42:
#281 renamed the environment variable for the python version of our snapshot CI test from `python` to `python-version`. The uv action could not find `env.python` and thus saw no input for `python-version`, which it needs to create a venv. And without a venv, we get [the error we see in today's snapshot CI test](https://github.com/iiasa/message-ix-models/actions/runs/12922591067/job/36038523512).

This PR adjust the name of the environment variable used for the uv action.

## How to review

- Read the diff and note that the variable name used now is correct.

## PR checklist


khaeru (approved):
Good catch, thanks!

PR #285, opened 2025-01-27
Make changes related to reduction year and pre-base year fix_cost in `tools.costs` 

measrainsey:
PR to change the year the reduction value is reached and to modify the pre-base year fix_cost in MESSAGEix scenario

I added a functionality to specify the year that the reduction rate is reached for the reference region. Currently the default is still 2150, but for ScenarioMIP it should be changed to 2150 (this can be specified in the `Config` object).

Additionally, I wrote some functions to update the `inv_cost` and `fix_cost` in a MESSAGEix Scenario object directly. The functions also modifies the inv_cost and fix_cost for pre-base year years (which are not projected in the costs tool). 

## How to review

Perhaps either @OFR-IIASA or @macflo8 can check if the changes are doing what is desired?


measrainsey:
For the sake of having the `ssp-dev` branch be up-to-date to have these changes in the next ScenarioMIP runs, I will go ahead and merge now

PR #286, opened 2025-01-27
Remove `final_year` attribute from Config

measrainsey:
The `final_year` attribute has been replaced by `final_model_year`. Currently `final_year` is still kept in the Config, but it should really be removed. In this PR, I would be removing the `final_year` attribute.

~However, this would cause some parts of the codebase to break. Namely, when `gen_te_projections()` is used, as that function has a `tools.costs` `Config` class that uses the `final_year` parameter:~

1. https://github.com/iiasa/message-ix-models/blob/ea8477ab40dca36dc7ed5777fc1b2fa414adcfb7/message_ix_models/project/ssp/script/util/functions.py#L281-L288
2. https://github.com/iiasa/message-ix-models/blob/ea8477ab40dca36dc7ed5777fc1b2fa414adcfb7/message_ix_models/project/ssp/script/scenarios/ccs_dac.py#L185-L192
3. https://github.com/iiasa/message-ix-models/blob/ea8477ab40dca36dc7ed5777fc1b2fa414adcfb7/message_ix_models/model/material/data_util.py#L1325-L1332

~One fix would be to update~ The PR also updates the cases above to use (the default value of) `final_model_year = 2110` instead of `final_year = 2110`. 

Another fix would be to switch to using the `update_scenario_costs()` function in `message_ix_models.tools.costs.scenario` that was developed as part of #285.

## How to review

~I think the people using the `tools.costs` Config object are @macflo8 and @ywpratama, so perhaps they have some ideas on preferences for moving forward.~

@macflo8 to confirm that the modified code does or will run without error.

### DIFF
--- a/message_ix_models/model/material/data_util.py
+++ b/message_ix_models/model/material/data_util.py
@@ -1347,7 +1347,6 @@ def gen_te_projections(
         method=method,
         format="message",
         scenario=ssp,
-        final_year=2110,
         reduction_year=reduction_year,
     )
     out_materials = create_cost_projections(cfg)
--- a/message_ix_models/model/material/data_util.py
+++ b/message_ix_models/model/material/data_util.py
@@ -326,9 +326,11 @@ def read_timeseries(
 
     if filename.endswith(".csv"):
         df = pd.read_csv(package_data_path("material", material, filename))
+
         # Function to convert string to integer if it is a digit
         def convert_if_digit(col_name):
             return int(col_name) if col_name.isdigit() else col_name
+
         # Apply the function to the DataFrame column names
         df = df.rename(columns=convert_if_digit)
     else:
@@ -1499,8 +1501,8 @@ def calibrate_t_d_tecs(scenario: "Scenario"):
     for i in td.index:
         try:
             row = td.iloc[i]
-        except:
-            here = 1
+        except Exception:
+            here = 1  # noqa: F841 —for debugging?
         # Retrieve input of techologies which take from the commodity/level.
         inp = scenario.par(
             "input",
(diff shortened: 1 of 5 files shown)


message_ix_models/model/material/data_util.py
```diff
@@ -1347,7 +1349,6 @@ def gen_te_projections(
         method=method,
         format="message",
         scenario=ssp,
-        final_year=2110,
```
khaeru, line 1350: FYI @macflo8.

message_ix_models/project/ssp/script/scenarios/ccs_dac.py
```diff
@@ -181,14 +192,15 @@ def gen_te_projections(
...
+    assert ssp in "all LED SSP1 SSP2 SSP3 SSP4 SSP5".split()
+
     model_tec_set = dac_techs
     cfg = Config(
         module="dac",
         ref_region=ref_reg,
         method=method,
         format="message",
-        scenario=ssp,
-        final_year=2110,
```
khaeru, line 191: FYI @ywpratama.

message_ix_models/project/ssp/script/util/functions.py
```diff
@@ -284,7 +284,6 @@ def gen_te_projections(
         method=method,
         format="message",
         scenario=ssp,
-        final_year=2110,
```
khaeru, line 287: FYI @ywpratama.

macflo8:
Just completed the test run on the `message_data` workflow mentioned in https://iiasa-ece.slack.com/archives/CD0GBHHA4/p1738856153627389?thread_ts=1738567802.232809&cid=CD0GBHHA4.
Can confirm that the changes do not break the workflow.

macflo8 (approved):

PR #259, opened 2024-12-03
Transport improvements from 2024-W49

khaeru:
- Items for SSP 2024/ScenarioMIP.
  - [x] Update base scenario URLs from e.g. `ixmp://ixmp-dev/SSP_dev_SSP1_v0.1_Blv0.18/baseline_prep_lu_bkp_solved_materials` to e.g. `ixmp://ixmp-dev/SSP_SSP1_v1.1/baseline_DEFAULT_step_13`.
  - [x] Constrain t=conm_ar (a P AIR technology) to a low/zero share in 2020. This technology was (a) taking a large share in y=2020 without constraints and (b) because not present at all in IEA EWEB for some regions, was being scaled with 1.0, so passing through as a large value.
  - ~[ ]~ Next PR: Use a smaller `growth_activity_up` value for (t=ICAm_ptrp, n=R12_RCPA). The technology constraint as of 2020 is effective, but this tech/region grows too rapidly to 2025 and beyond. (per discussion with @r-aneeque 2024-12-04.)
  - ~[ ]~ Next PR: Adjust efficiencies of LDV PHEV technologies using duty cycle data.
  - ~[ ]~ Next PR: Ensure `growth_activity_up` and `act-non_ldv.csv` values are set for `F ROAD` technologies with, at least, c=gas inputs. This is to avoid the behaviour (observed by @r-aneeque) that the model quickly switches into these in 2025–2030. Rework `.transport.Config.constraint` to simplify the setting of these constraints.
- Items for aviation emissions postprocessing:
  - [x] Implement method 'B'.
  - [x] #230

This branch ~is currently~ was temporarily triggering the message_data transport runs using iiasa/message_data#582, which limits to a subset of all scenarios.

## How to review

Per aviation emissions:
- View the diff (click 'Files changed' above), particularly the file `message_ix_models/data/transport/emi-intensity.csv`.
- View the [**preview build**](https://iiasa-energy-program-message-ix--259.com.readthedocs.build/projects/models/en/259/) of the documentation, in particular the page “Shared Socioeconomic Pathways (project.ssp)” and section “2024 update > transport”, [here](https://iiasa-energy-program-message-ix--259.com.readthedocs.build/projects/models/en/259/project/ssp.html#update).
[...]

PR #289, opened 2025-01-29
Transport improvements from 2025-W05

khaeru:
Carried from #259:
- ~Use a smaller `growth_activity_up` value for (t=ICAm_ptrp, n=R12_RCPA). The technology constraint as of 2020 is effective, but this tech/region grows too rapidly to 2025 and beyond. (per discussion with @r-aneeque 2024-12-04.)~ Next PR.
- ~Adjust efficiencies of LDV PHEV technologies using duty cycle data.~ Next PR.
- [x] Ensure `growth_activity_up` and `act-non_ldv.csv` values are set for `F ROAD` technologies with, at least, c=gas inputs. This is to avoid the behaviour (observed by @r-aneeque) that the model quickly switches into these in 2025–2030.
- ~Rework `.transport.Config.constraint` to simplify the setting of these constraints.~ Next PR.

In addition:
- [x] Adjust transport code and tests for genno 1.28.x. This fixes failures of scheduled jobs of the "pytest" workflow on `main`.
- [x] Drop support for ixmp/message_ix 3.6.0, in line with the [upstream version policy](https://docs.messageix.org/projects/models/en/latest/develop.html#upstream-version-policy).
- [x] Establish SDMX-CSV format for transport outputs for EDITS MCE.
- [x] Improve the dispatch of the "transport" workflow in message_data (see also iiasa/message_data#596). With this, the set of scenarios run can be adjusted by making (temporary or permanent) changes in transport.yaml in *this* repository. This reduces the need for adjustments on the message_data side.

## How to review

- Read the additions to doc/whatsnew.
- Note the CI checks all pass.

## PR checklist

- [x] Update doc/whatsnew.
- [x] After review approval: rebase to drop the "TEMPORARY" commits.


khaeru:
> Could it be that we're now running the pytest workflow file from `main` again, which is still trying to run the v3.6.0 tests, even though this PR dropped support for them?

Yes, exactly so.

- The PR contains commits that modify the pytest.yaml workflow file, e.g. "Drop support for ixmp/message-ix version 3.6"
- However, because the default setting is to run that workflow on the 'pull_request_target' event, the version of the workflow file used is the one on `main`, not the modified one on the branch.
- In order to 'prove' that all jobs will succeed once once the changed pytest.yaml is merged back to `main`, I add a TEMPORARY commit that turns off the 'pull_request_target' and turns on the 'pull_request' trigger. This results in the workflow file from the branch being used, and we saw that the jobs succeeded.
- In order to avoid disturbing 'pull_request_trigger' config on `main`, we need to drop that TEMPORARY commit before merging. When we do, the `main` version of the workflow will run again, and will fail.

This is an unfortunate complication of our setup using 'pull_request_trigger' to allow use of secrets on PRs from forks. It is only necessary when modifying the workflow file itself. Maybe there could be some clever way to circumvent the need to do it this way, but I haven't had time to consider that yet. At minimum I guess I should write down somewhere that this process is the one we must use.

PR #291, opened 2025-02-05
Add `.util.slurm` and CLI

### LINKED ISSUE #284: Add `.util.slurm` and CLI
Building on #279, some simple utility code could reduce the difficulty of using message-ix-models with SLURM on UniCC and elsewhere. I imagine the following as minimum functionality:

- The user runs a command like `mix-models sbatch [options] -- transport run --nodes=R12 --from="SSP2 built" "SSP2 solved"` in which:
  - `mix-models sbatch` is the entry point.
  - `[options]` control the interaction with SLURM/sbatch.
  - `transport run …` are the commands to be run.
- The code generates (maybe using a Jinja2 template) a job file in the user's cache/temporary working directory that resembles:
  ```
  #!/bin/sh
  #SBATCH --time=3:00:00
  #SBATCH --mem=40G
  #SBATCH --mail-type=BEGIN,END,FAIL
  #SBATCH --mail-user=username@iiasa.ac.at
  #SBATCH -o ~/out/solve_%J.out
  #SBATCH -e ~/err/solve_%J.err
  
  module purge
  source /opt/apps/lmod/8.7/init/bash
  module load Python/3.11.5-GCCcore-13.2.0
  module load Java

  uv venv …    
    
  mix-models transport run --nodes=R12 --from="SSP2 built" "SSP2 solved
  ```
  …in which:

  - The `#SBATCH` lines are assembled from configuration; the CLI `[options]`; and common defaults.
  - The lines like `module purge` and other preamble are from configuration.
  - The `uv venv` line creates a node-specific virtual environment and installs message-ix-models and its dependencies. This could also work by activating a particular virtual environment know to already exist on the node; this would reduce start-up time but also risk of file collisions etc.
  - The line `mix-models transport run …` exactly matches the invocation on the command line.
- The code invokes `sbatch job.sh` and returns any output to the user.

This could later be extended to, for instance, run multiple tasks like `#SBATCH --ntasks=10` / `srun --ntasks=1 mix-models transport …`. This would require generating a list of individual tasks.

khaeru:
This PR…
- Adds a `mix-models sbatch` CLI command that simplifies usage with SLURM, for instance on UnICC.
  This closes #284.
- Collects some docs pages under a 'HOWTO' section, with instructions on HOWTO Write a HOWTO.

## How to review

- Read the [added documentation](https://iiasa-energy-program-message-ix--291.com.readthedocs.build/projects/models/en/291/api/util.html#module-message_ix_models.util.slurm) via the preview build.
- Check out the branch on UnICC.
- Run a command like `mix-models sbatch -- config show`.

## PR checklist

- [x] Update doc/whatsnew.

### DIFF
--- a/doc/api/util.rst
+++ b/doc/api/util.rst
@@ -225,6 +225,14 @@ Commonly used:
 .. automodule:: message_ix_models.util.sdmx
    :members:
 
+:mod:`.util.slurm`
+==================
+
+.. currentmodule:: message_ix_models.util.slurm
+
+.. automodule:: message_ix_models.util.slurm
+   :members:
+
 :mod:`.types`
 =============
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -211,6 +211,7 @@ intersphinx_mapping = {
     "pytest": ("https://docs.pytest.org/en/stable/", None),
     "python": ("https://docs.python.org/3/", None),
     "sdmx": ("https://sdmx1.readthedocs.io/en/stable/", None),
+    "sphinx": ("https://www.sphinx-doc.org/en/master", None),
 }
 
 # -- Options for sphinx.ext.linkcode / ixmp.util.sphinx_linkcode_github ----------------
--- a/doc/distrib.rst
+++ b/doc/distrib.rst
@@ -1,5 +1,5 @@
-Overview and tooling
-********************
+Distributed computing
+*********************
 
 This page introduces considerations, tools, and features for using **distributed** or **high-throughput computing** with MESSAGEix-GLOBIOM.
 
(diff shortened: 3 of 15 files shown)


doc/api/util.rst
glatterf42, line 1: The docs-rendering of the `TEMPLATE` variable does not seem particularly helpful. I'm not sure if we want that to include the full template or if we're fine without (e.g. only linking the template), in which case we could maybe exclude that variable.

glatterf42 (approved):
Thanks, looks good to me in general. Just one question inline and the note that I haven't checked out anything on UniCC, as I'm not sure I can connect there without asking ICT for permission first. 
Or is that where we run jupyterhub, too? If so, I could open a terminal in that application and try.

doc/api/util.rst
khaeru: Agreed. One of the enhancements (that I chose to leave out of this PR) will be to remove TEMPLATE altogether and build it up from various different pieces, e.g. the SBATCH options, the environment initialization, etc.

At that point I think we can replace the variable display in the docs with an example of what a fully-constructed job script looks like.

PR #295, opened 2025-02-12
Confirm support for minimum versions

### LINKED ISSUE #294: Two issues with Python ≤ 3.10, sdmx < 2.16
@jkikstra reports two issues using older dependencies, namely with Python 3.10 and sdmx 2.15:

1. https://github.com/iiasa/message-ix-models/blob/43ce610356d84d6134936493e0226b6ee2f71ad4/message_ix_models/types.py#L40

   …raises `AttributeError: module 'sdmx.model.common' has no attribute 'Version'`. This should be resolved by updating the minimum version here to [2.16](https://sdmx1.readthedocs.io/en/latest/whatsnew.html#v2-16-0-2024-08-16):

   https://github.com/iiasa/message-ix-models/blob/43ce610356d84d6134936493e0226b6ee2f71ad4/pyproject.toml#L46

   and/or enclosing the type annotation in a string.

2. https://github.com/iiasa/message-ix-models/blob/43ce610356d84d6134936493e0226b6ee2f71ad4/message_ix_models/util/cache.py#L50

   raises: `TypeError: Invalid annotation for 'o'. typing.Union[genno.core.attrseries.AttrSeries, genno.core.sparsedataarray.SparseDataArray] is not a class.`. This is only supported from Python 3.11. This should be resolved, first by updating the pytest matrix to cover the declared supported versions, e.g. using Python 3.9 here: https://github.com/iiasa/message-ix-models/blob/43ce610356d84d6134936493e0226b6ee2f71ad4/.github/workflows/pytest.yaml#L77

   …and then, for Python ≤ 3.10, making distinct registrations for the union-ed types.

khaeru:
Closes #294. Supersedes #293.

## How to review

- @glatterf42: Read the diff and note that the CI checks all pass.
- @macflo8 @r-aneeque: You may have got an automatic notification as code owners, but I've removed the request: There is no functional change to code, only to type hints in order to support this older version of Python.

## PR checklist

- ~Add, expand, or update documentation.~ N/A; CI and type hint changes only
- [x] Update doc/whatsnew.
- [x] **After review,** rebase the PR to drop the 'TEMPORARY' commit to run the workflow on the pull_request rather than pull_request_target event.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -74,7 +74,8 @@ jobs:
         # - Versions of ixmp and message_ix to test.
         # - Latest supported Python version for those or other dependencies.
         # Minimum version given in pyproject.toml + earlier version of Python
-        - { upstream: v3.6.0, python: "3.11" }  # 2022-08-18
+        # For this job only, the oldest version of Python supported by message-ix-models
+        - { upstream: v3.6.0, python: "3.9" }  # Released 2022-08-18
         - { upstream: v3.7.0, python: "3.11" }  # 2023-05-17
         - { upstream: v3.8.0, python: "3.12" }  # 2024-01-12
         # Latest released version + latest released Python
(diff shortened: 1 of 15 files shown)


khaeru:
Great, thanks. Since we don't expect the CI to pass after this rebase (because the pull_request_trigger operates on the base branch), I will merge without waiting for it to complete.

PR #299, opened 2025-02-24
Test against ixmp/message-ix v3.10.0

khaeru:
With ixmp/message-ix v3.10.0 both released, we add these to the "pytest" CI workflow matrix. This will help us ensure that message-ix-models continues to work with both these latest releases and the evolving `main` branches for those repos.

According to our [Upstream version policy](https://docs.messageix.org/projects/models/en/latest/develop.html#upstream-version-policy):
> [message-ix-models supports t]he most recent 4 minor versions [of ixmp/message-ix], or all minor versions released in the past two (2) years—whichever is greater.

- The most recent 4 minor versions = 3.7, 3.8, 3.9, 3.10.
- Versions released in the last 2 years = versions released on 2023-02-24 or later. [v3.7.0 is the oldest version released since that date (on 2023-05-17)](https://pypi.org/project/message-ix/#history).

In this case both criteria give the same minimum version. Support and CI jobs for upstream 3.6 were already dropped in #289.

## How to review

None required.

Note that because of our use of 'pull_request_trigger', the added jobs will not run until this branch is merged. However, I don't believe we need the extra effort of showing they run on this branch, since at this moment upstream `v3.10.0` and `main` are identical.

## PR checklist

- [x] Continuous integration checks all ✅
- [x] ~Add or expand tests;~ coverage checks both ✅
- ~Add, expand, or update documentation.~
- [x] Update doc/whatsnew.

PR #302, opened 2025-02-25
Fix CI for `message_ix` v3.10.0

glatterf42:
The recent release v3.10.0 of message_ix and ixmp and #299 broke our CI by exposing that we don't properly handle version comparisons for extra comparisons at the moment. This PR fixes that.

In our `pytest` CI workflow, we add extra dependency requirement for specific versions. At the moment, this is done based on string comparison, but with `"v3.10.0"` coming into play, these begin to fail: 

```python
>>> "v3.10.0" < "v3.1.0"
False
>>> "v3.10.0" < "v3.7.0"
True
``` 

This results in incompatible dependencies being installed, causing the CI to fail.
@khaeru suggested to use [packaging.version](https://packaging.pypa.io/en/latest/version.html) instead of string comparisons, which I tried to add now. There's just one catch: these versions can't parse `"main"`. 
I thought about using an arbitrary number in that case that's high enough to always trigger the checks for the latest version, e.g. `99.99.99`. However, I don't really like such magic numbers and eventually we might reach that version. Also, there's part two of this PR:

One of the checks we have targets the latest versions of this repo. This should have been handled by #225, though, so this PR removes the workaround. And I generally think we want to avoid version pins for the `"main"` version, so right now, the dependency additions are hidden behind an `if version_string != "main"` check. 

## How to review

## PR checklist

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -129,17 +129,20 @@ jobs:
       id: dependencies
       run : |
         from os import environ
+        from packaging.version import Version, parse
         from pathlib import Path
 
-        v, result = "${{ matrix.version.upstream }}".replace("main", "vmain"), []
-        for condition, dependency in (
-            (v >= "v3.7.0", "dask[dataframe] < 2024.11.0"),  # dask >= 2024.11.0 changes handling of dict (will be addressed in #225)
-            (v <= "v3.7.0", "genno < 1.25"),  # Upstream versions < 3.8.0 import genno.computations, removed in 1.25.0 (#156)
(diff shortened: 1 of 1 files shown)


.github/workflows/pytest.yaml
```diff
@@ -125,21 +125,27 @@ jobs:
       # Work around ts-graphviz/setup-graphviz#630
       if: matrix.os != 'macos-13'
 
+    - name: Install packaging
+      run: uv pip install packaging
```
glatterf42, line 129: I thought `packaging` was a built-in module, but without this step, we get a `ModuleNotFound` below.

khaeru: Unfortunately it's not, but it's provided by the PSF themselves, so hopefully can be regarded as stable and low-risk.

PR #320, opened 2025-03-26
Fix Code Quality CI: checkout action and complex function

glatterf42:
This PR intends to fix a few things about the Code Quality CI:

- It fixes a typo such that the Code Quality check will consider the changes here, not those already on `main`
- It reduces the complexity of the `cool_tech()` function to below 11, which is our accepted limit here (@adrivinca, this may serve as a template for the future, if you're interested: I have simply extracted one `if/else` clause and moved that to its own function, following the idea that every function should have one clearly defined task)
- I've also deleted the `pre-commit` cache in hopes of it now picking up the latest version of genno, which should fix the mypy errors we've been seeing in the nightly tests

Housekeeping:
- Mirror the 'fix' from khaeru/genno#166 for a mypy error.
- Adjust import of the Annotation class from sdmx1 to avoid warnings added in v2.22.

## How to review

## PR checklist

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -189,7 +189,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
       with:
-        ref: ${{ needs.check.output.ref }}
+        ref: ${{ needs.check.outputs.ref }}
     - uses: astral-sh/setup-uv@v5
       with:
         cache-dependency-glob: "**/pyproject.toml"
--- a/message_ix_models/model/bare.py
+++ b/message_ix_models/model/bare.py
@@ -115,7 +115,7 @@ def get_spec(context) -> Spec:
     nodes = get_codes(f"node/{context.model.regions}")
 
     # Top-level "World" node
-    world = nodes[nodes.index("World")]
+    world = nodes[nodes.index(Code(id="World"))]
 
     # Set elements: World, followed by the direct children of World
     add.set["node"] = [world] + world.child
--- a/message_ix_models/model/disutility.py
+++ b/message_ix_models/model/disutility.py
@@ -7,7 +7,8 @@ from itertools import product
 
 import message_ix
 import pandas as pd
-from sdmx.model.common import Annotation, Code
+from sdmx.model.common import Code
+from sdmx.model.v21 import Annotation
 
 from message_ix_models import ScenarioInfo, Spec
 from message_ix_models.model.build import apply_spec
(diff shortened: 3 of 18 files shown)


glatterf42:
I have also deleted the cache entry called `setup-uv-1-x86_64-unknown-linux-gnu-3.9-...`, as this is used by the setup-uv action in the code quality run, but it doesn't seem to have any effect. 
Also, even though the linked [workflow file](https://github.com/iiasa/message-ix-models/actions/runs/14080388674/workflow?pr=320) uses `needs.check.outputs.ref` for the code quality check now, the checkout action still resolves to a0d542014d8a6908410137ec6662202c7bf8cf66, the latest commit on `main`.

I'll have to do some organizing for the Community Meeting now and can only come back to this later, I'm afraid.

khaeru:
> even though the linked [workflow file](https://github.com/iiasa/message-ix-models/actions/runs/14080388674/workflow?pr=320) uses `needs.check.outputs.ref` for the code quality check now, the checkout action still resolves to [a0d5420](https://github.com/iiasa/message-ix-models/commit/a0d542014d8a6908410137ec6662202c7bf8cf66), the latest commit on `main`.

I think that is expected. The web display on GitHub itself seems to be erroneous; the `pull_request_target` trigger always uses the version of the workflow file on the target branch, in this case `main`.

In order to debug, you can do what it says here:
https://github.com/iiasa/message-ix-models/blob/a0d542014d8a6908410137ec6662202c7bf8cf66/.github/workflows/pytest.yaml#L4-L9
in a commit with a message like "TEMPORARY Run 'pytest' workflow from PR branch"; preserve a link to the successful run; and then rebase the branch to drop the temporary commit.

glatterf42:
Thanks, I had forgotten that was necessary.

adrivinca:
Thanks, looks good to me. should we write a test for the new function?

glatterf42:
Even adding the main branch of genno to the additional requirement for mypy does not seem to resolve the errors. I don't really understand what's going on there as all these type hints are very abstract. I could dive in, but I think I should be doing other things instead. So I would opt for applying `type: ignore` if we have tests for both places where mypy raises errors and these pass.

Unfortunately, there is now another error originating from the latest release of sdmx1, and I don't really have time to look into that, either :/

khaeru:
> Unfortunately, there is now another error originating from the latest release of sdmx1, and I don't really have time to look into that, either :/

That's something I caused, so I'll step in with a commit to fix it.

khaeru:
> I'll step in with a commit to fix it.

I believe I've addressed the issues related to genno and sdmx1, so back to you @glatterf42.

glatterf42:
These tests will fail again, at least in part, because they're using the code stored on `main`. Since they were passing before, I'll merge without waiting for them.

Issue #314, opened 2025-03-21
WEO_2023_PG_Assumptions_STEPSandNZE_Scenario.xlsx Not downloading properly from pip


Tyler-lc:
Also, once you have it working properly, this happens:
```
[c:\***\***message-ix-models\message_ix_models\tools\costs\regional_differentiation.py:125](file:///C:/vs_code/hyway_ssp/message-ix-models/message_ix_models/tools/costs/regional_differentiation.py:125): FutureWarning: Downcasting behavior in `replace` is deprecated and will be removed in a future version. To retain the old behavior, explicitly call `result.infer_objects(copy=False)`. To opt-in to the future behavior, set `pd.set_option('future.no_silent_downcasting', True)
```

I think this is a relatively new warning from pandas. But it is worth looking into it I guess.

khaeru:
A couple things about tracking issues:
- Parent/sub-issues should be used only when there is an actual relationship between two issues.

  In this case, the current issue/warning comes from the module `.tools.costs.regional_differentiation`. This happens to be the same module (and function) that reads the file mentioned in #313. However, as you describe, this warning occurs *even if* issue 313 is solved/not encountered. So while you may have encountered this issue in the course of (or after) investigating the other one, they seem independent.

  In other words, fixing this issue is not a necessary or sufficient condition for fixing issue 313. For that reason, it should not be a sub-issue.
- A good, explanatory title can help.

  In this case, a title that describes the issue *itself* (and not the parent that, per the above point, should not be the parent in any case) might be: **FutureWarning from pandas in `get_weo_data()`**. This describes what is happening and where.

PR #322, opened 2025-03-30
Adjust packaging of Git LFS artefacts

### LINKED ISSUE #313: ValueError in `get_weo_data()` from PyPI distribution due to Git LFS pointers
I have been trying to use the `create_cost_projections` method using the [demo.py](https://github.com/iiasa/message-ix-models/blob/main/message_ix_models/tools/costs/demo.py).
The problem is that for some reason some of the files are not downloading properly when installing with pip. At this point I was using message 3.9 together with ixmp 3.9 and message_ix_models: '2024.8.6'
When I try to run `res_r12_energy = create_cost_projections(cfg)`, i receive an error saying: 

```---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[2], line 3
      1 cfg = Config()
----> 3 res_r12_energy = create_cost_projections(cfg)

File c:\Users\c******\AppData\Local\anaconda3\envs\message310\Lib\site-packages\message_ix_models\tools\costs\projections.py:662, in create_cost_projections(config)
    655 func = {
    656     "convergence": create_projections_converge,
    657     "gdp": create_projections_gdp,
    658     "constant": create_projections_constant,
    659 }[config.method]
    661 # Create projections
--> 662 df_costs = func(config)
    664 # Convert to MESSAGEix format
    665 df_inv, df_fom = create_message_outputs(df_costs, config)

File c:\Users\c******\AppData\Local\anaconda3\envs\message310\Lib\site-packages\message_ix_models\tools\costs\projections.py:153, in create_projections_gdp(config)
    150 log.info(f"Selected scenario version: {config.scenario_version}")
    152 log.info("Calculate regional differentiation in base year+region")
--> 153 df_region_diff = apply_regional_differentiation(config)
    155 log.info("Apply cost reduction rates to reference region")
    156 df_ref_reg_reduction = project_ref_region_inv_costs_using_reduction_rates(
    157     df_region_diff, config
    158 ).pipe(_maybe_query_scenario, config)

File c:\Users\c******\AppData\Local\anaconda3\envs\message310\Lib\site-packages\message_ix_models\tools\costs\regional_differentiation.py:678, in apply_regional_differentiation(config)
    676 df_map = adjust_technology_mapping(config.module)
    677 assert config.ref_region is not None
--> 678 df_weo = get_weo_regional_differentiation(config)
    679 df_intratec = get_intratec_regional_differentiation(config.node, config.ref_region)
    681 # Get mapping of technologies
    682 # Then merge with output of get_weo_regional_differentiation
    683 # If the base_year_reference_region_cost is empty, then use the weo_ref_region_cost
    684 # If the fix_ratio is empty, then use weo_fix_ratio

File c:\Users\c*****\AppData\Local\anaconda3\envs\message310\Lib\site-packages\message_ix_models\tools\costs\regional_differentiation.py:476, in get_weo_regional_differentiation(config)
    449 """Apply WEO regional differentiation.
    450 
    451 1. Retrieve WEO data using :func:`.get_weo_data`.
   (...)    472     - reg_cost_ratio: regional cost ratio relative to reference region
    473 """
    475 # Grab WEO data and keep only investment costs
--> 476 df_weo = get_weo_data()
    478 # Even if config.base_year is greater than 2022, use 2022 WEO values
    479 sel_year = str(2022)

File c:\Users\c****\AppData\Local\anaconda3\envs\message310\Lib\site-packages\message_ix_models\tools\costs\regional_differentiation.py:99, in get_weo_data()
     96 dfs_cost = []
     97 for tech_key, cost_key in product(DICT_TECH_ROWS, DICT_COST_COLS):
     98     df = (
---> 99         pd.read_excel(
    100             file_path,
    101             sheet_name=DICT_TECH_ROWS[tech_key][0],
    102             header=None,
    103             skiprows=DICT_TECH_ROWS[tech_key][1],
    104             nrows=9,
    105             usecols=DICT_COST_COLS[cost_key],
    106         )
    107         .set_axis(["weo_region", "2022", "2030", "2050"], axis=1)
    108         .melt(id_vars=["weo_region"], var_name="year", value_name="value")
    109         .assign(
    110             weo_technology=tech_key,
    111             cost_type=cost_key,
    112             units="usd_per_kw",
    113         )
    114         .reindex(
    115             [
    116                 "cost_type",
    117                 "weo_technology",
    118                 "weo_region",
    119                 "year",
    120                 "units",
    121                 "value",
    122             ],
    123             axis=1,
    124         )
    125         .replace({"value": "n.a."}, np.nan)
    126         .assign(value=lambda x: x.value * conversion_factor)
    127     )
    129     dfs_cost.append(df)
    131 all_cost_df = pd.concat(dfs_cost)

File c:\Users\c******\AppData\Local\anaconda3\envs\message310\Lib\site-packages\pandas\io\excel\_base.py:495, in read_excel(io, sheet_name, header, names, index_col, usecols, dtype, engine, converters, true_values, false_values, skiprows, nrows, na_values, keep_default_na, na_filter, verbose, parse_dates, date_parser, date_format, thousands, decimal, comment, skipfooter, storage_options, dtype_backend, engine_kwargs)
    493 if not isinstance(io, ExcelFile):
    494     should_close = True
--> 495     io = ExcelFile(
    496         io,
    497         storage_options=storage_options,
    498         engine=engine,
    499         engine_kwargs=engine_kwargs,
    500     )
    501 elif engine and engine != io.engine:
    502     raise ValueError(
    503         "Engine should not be specified when passing "
    504         "an ExcelFile - ExcelFile already has the engine set"
    505     )

File c:\Users\c******\AppData\Local\anaconda3\envs\message310\Lib\site-packages\pandas\io\excel\_base.py:1554, in ExcelFile.__init__(self, path_or_buffer, engine, storage_options, engine_kwargs)
   1550     ext = inspect_excel_format(
   1551         content_or_path=path_or_buffer, storage_options=storage_options
   1552     )
   1553     if ext is None:
-> 1554         raise ValueError(
   1555             "Excel file format cannot be determined, you must specify "
   1556             "an engine manually."
   1557         )
   1559 engine = config.get_option(f"io.excel.{ext}.reader", silent=True)
   1560 if engine == "auto":

ValueError: Excel file format cannot be determined, you must specify an engine manually.
```

I tracked the path that the function was trying to point to, and I found that, when using pip install [this file](https://github.com/iiasa/message-ix-models/blob/main/message_ix_models/data/iea/WEO_2023_PG_Assumptions_STEPSandNZE_Scenario.xlsx) is not downloading properly and it is only 1kB, as opposed to the 65KB expected.
So I tried to create a new environment, fresh. All new installation with the latest packages (message_ix: 3.10.0, ixmp: 3.10.0, message_ix_models: 2025.1.10). Still the same issue.

I tried to manually overwriting that single file, that I manually downloaded from the main repo. But now, I get a new error, because another file is broken. This is where I stop tracking all the files :) 

Then I cloned the repo inside of a local project. Meaning that I have the following folder structure:
| > personal_project
|      | > my_utils
|      | my_scripts.py
|      | > message-ix-models
|      |_______| > *mesage-ix-models folders*

This method still needs some work to function properly. If you just try to run the message-ix-models as a local package, you will find some issues with dependencies. 

So, what worked:
1. Cloning the repo and running a script from inside the message-ix-models folder. Trying to run the script from outside of it, importing the message-ix-models as a package does not work natively.
2. I want to run a script from outside the message-ix-models folder (because I am just using the packages really, I do not want -yet- to modify things in there). For this reason I also tried (and it worked) to `run pip install -e .` from inside the message-ix-folder. But then you also have to set the `sys.path.append(os.path.abspath("./message-ix-models")` because you can't import something that has a hyphen. 

In any case, when cloning the repo locally, all the files are present and are not broken. Using pip brakes some files. At least two. I did not try to replace all files manually until everything worked. Which I could do now that I cloned the repo, just replace the data folder in the anaconda packages folder. But that would obviously not be the intended way.
I hope the lengthy explanation is helpful to you! 

Cheers,
L.

khaeru:
This PR closes #313 by using the `lfs: true` input to the reusable publish.yaml workflow, added by iiasa/actions#14.

Also:
- Expand docs of `.tools.costs` to describe the required SSP data.
- Housekeeping: address this warning from `python -m build`:
  ```
  SetuptoolsDeprecationWarning: License classifiers are deprecated.
  !!
        ********************************************************************************
        Please consider removing the following classifiers in favor of a SPDX license expression:

        License :: OSI Approved :: Apache Software License

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************
  !!
  ```
## How to review

- Check the operation of the publish.yaml workflow. In particular:
  - The Git tag `v2025.3.30a0` was created and pushed on this branch.
  - That resulted in [this workflow run](https://github.com/iiasa/message-ix-models/actions/runs/14160209282/job/39664619438#step:6:183).
  - [These files](https://pypi.org/project/message-ix-models/2025.3.30a0/#files) were published to PyPI.
  - In these, the file mentioned in #313 is included.
- Read the expanded [“Usage” section of the `.tools.costs` docs](https://iiasa-energy-program-message-ix--322.com.readthedocs.build/projects/models/en/322/api/tools-costs.html#usage), including following the links and instructions on how to obtain the necessary data.
- Install this version of the package (e.g. `uv pip install "message-ix-models==2025.3.30a0"`) and run the example code from #313 to confirm it works with the contents distributed via PyPI.

## PR checklist

- ~Add, expand, or update documentation.~ N/A, CI changes only.
- ~Update doc/whatsnew.~
- After approval:
  - [x] Merge iiasa/actions#14.
  - [x] Drop the TEMPORARY commit from this branch.
  - [x] Delete the `v2025.3.30a0` tag.


glatterf42 (commented):
Thanks for this fix :)
I can confirm the first several points in the review steps are working. Unfortunately, trying to run the code in the [demo.py](https://github.com/iiasa/message-ix-models/blob/main/message_ix_models/tools/costs/demo.py) did not work for me:
First, I discovered that when I'm setting up a venv in a directory that includes `message_data` as a subdirectory and I start a `python` terminal there, I can say `import message_data` without raising an error that we catch, but this still results in a `TypeError` because we [then](https://github.com/iiasa/message-ix-models/blob/main/message_ix_models/util/common.py#L23) want to access `Path(message_data.__file__)`, where `message_data.__file__` is `None`. This may not matter often, but it's an edge case we don't cover right now.
Changing to a directory without `message_data` in it fixed this. But `from message_ix_models.tools.costs.config import Config` still didn't work because plotnine was missing: `ModuleNotFoundError: No module named 'plotnine'`. So this seems like we need a new `tools/costs` section in `pyproject.toml` to ensure that these additional dependencies are present when required. Installing it manually fixed that error.
Now, I can run

```python
from message_ix_models.tools.costs.config import Config
from message_ix_models.tools.costs.projections import create_cost_projections
cfg = Config()
```

without errors in my fresh venv, but adding

```python
res_r12_energy = create_cost_projections(cfg)
```

causes this error:

```python
FileNotFoundError: [Errno 2] No such file or directory: '/home/fridolin/.virtualenvs/fix-git-lfs-pointers/lib/python3.12/site-packages/message_ix_models/data/ssp/1706548837040-ssp_basic_drivers_release_3.0_full.csv.gz'
```

I'm not sure where this file should come from; the files on PyPI don't include an `ssp` subdirectory in `message_ix_models/data/`. My clone that I'm using for work on message-ix-models otherwise does include that directory and file, but I don't recall creating them. What should we do here?

Tyler-lc:
I tried to install the version you indicated of message-ix-models in a fresh conda environment with the latest message-ix. And now it seems that those files that were broken before are there. Because after running the 
```
cfg = Config()
res_r12_energy = create_cost_projections(cfg)
```
I also checked manually if the file WEO_2023_PG_Assumptions_STEPSandNZE_Scenario.xlsx was there, and it is.
I receive the Pandas' warning about downcasting. So something is happening in that direction.
But I also a similar error as the one @glatterf42 is talking about.
And I can see that the folder /data/ssp is completely missing from the package's folder.

khaeru:
> I discovered that when I'm setting up a venv…

Hm, I see. Indeed, creating a directory (an empty one?) that shadows the name of *any* dependency is likely to create problems. There are some things to clarify here, but can this be a separate issue?

> `from message_ix_models.tools.costs.config import Config` still didn't work because plotnine was missing: `ModuleNotFoundError: No module named 'plotnine'.`

If we look at the the file in question, we can see that plotnine is not imported in that file. Can you please give the full traceback so we can see precisely where this exception is raised?

> FileNotFoundError: … 1706548837040-ssp_basic_drivers_release_3.0_full.csv.gz'

(Again it would help to see more of the traceback to know where this is happening, which would make it easier to diagnose.)

Although the SSP Scenario Explorer landing page ([here](https://data.ece.iiasa.ac.at/ssp/#/login)) says “publicly available” and the linked [license](https://data.ece.iiasa.ac.at/ssp/#/license) says “This Public License is adapted from the Creative Commons … License,” those ‘adaptations’ include restrictive terms that prohibit us from redistributing the source files. (The [actual Creative Commons licenses](https://en.wikipedia.org/wiki/Creative_Commons_license#Types_of_licenses) are defined by the absence of such restrictions.)

The file is stored in this repo using Git LFS, which is why you may see it in a separate checkout. To comply with the license, it is not packaged, but perhaps even having it as LFS in this public repo could be considered not compliant with the restrictive license. We should likely move it to https://github.com/iiasa/message-static-data. In any case, this falls under the "documentation" aspect I mentioned at https://github.com/iiasa/message-ix-models/issues/313#issuecomment-2750870713: because the license prevents us from distributing it to them, the user needs to get ahold of this file on their own.

So I will, for now, update the documentation to add these instructions; and we can separately address where to store a copy of the file for CI and other automated purposes.

khaeru:
> I receive the Pandas' warning about downcasting. So something is happening in that direction.

Please see https://github.com/iiasa/message-ix-models/issues/314#issuecomment-2750819521. To put that more clearly, this is a warning, not an error message: it is not clear it is the cause of the issue #313 that this PR attempts to address. I suspect it is not, and that it is only seen because the `.costs` code happens to be using that pandas functionality elsewhere/nearby. If that's correct, it should be a separate issue (not a sub-issue), and someone can address it through a separate PR.

Tyler-lc:
I am pretty sure that the Warning from Pandas is not the reason of the error we get from this PR. Because I get the same warning even when I manage to make everything work. If I remember correctly, the warning has to do with Pandas not inferring anymore the data type of the column automatically. And one would have to either manually tell Pandas to infer the type or (probably better) tell Pandas what the data type should be using astype(float) (for example).

khaeru:
> I am pretty sure that the Warning from Pandas is not the reason of the error we get from this PR.

Okay, we are agreed then. So please:
- Go to #314.
- In the right sidebar, under "Relationships / Parent issue", click the ⚙️ icon and choose "Change or remove parent"
- Remove 313 as the parent issue.
- Edit the title of 314 so it describes the distinct thing to be addressed, rather than repeating the original title of 313.

khaeru:
> So I will, for now, update the documentation to add these instructions; and we can separately address where to store a copy of the file for CI and other automated purposes.

@glatterf42 @Tyler-lc please see the added second bullet under “How to review”.

glatterf42:
Thanks, I think the instructions are clear now. I would only add links to where people should register to the [data section](https://iiasa-energy-program-message-ix--322.com.readthedocs.build/projects/models/en/322/project/ssp.html#module-message_ix_models.project.ssp.data), so I think we can e.g. link `https://data.ece.iiasa.ac.at/ssp` (and maybe another one for 2017/2024) on the page to make it easier for people to register at the correct location.

Agreed that the others should be separate issues, that's part of the reason I didn't post the whole tracebacks immediately. I'll open them now :)

glatterf42 (approved):
Thanks, looks good to me now :)

The links to the data are not at the top of the page, but we may want anyone using the data to read the docs first, anyway :)

khaeru:
@Tyler-lc since you reported #313, I would ultimately appreciate your ✅ if the added docs do indeed help get the code to work on your system.

Tyler-lc:
I think it is clear enough, for a somewhat experienced user. But I'm wondering whether the files that we have up in the repo, are "fuzzed" data or not. Because, honestly, I can't tell if those that we have publicly available on the repo are real or not. Does anyone know?

Tyler-lc:
Also, I am not sure where the submission form would be. And where to find the downloadable real data. At least not from the documentation. Where would one find the submission form and the data?

khaeru:
Thanks for the comments.

> Also, I am not sure where the submission form would be. And where to find the downloadable real data. At least not from the documentation. Where would one find the submission form and the data?

By https://github.com/iiasa/message-ix-models/pull/322#issuecomment-2765846396 I was referring to [this line](https://github.com/iiasa/message-ix-models/blame/2619da07c3e9581cedfc56e669ba285472437bfc/message_ix_models/project/ssp/data.py#L21) and [this line](https://github.com/iiasa/message-ix-models/blame/2619da07c3e9581cedfc56e669ba285472437bfc/message_ix_models/project/ssp/data.py#L125) that I added in this PR. In the docs, those URLs appear respectively [here](https://iiasa-energy-program-message-ix--322.com.readthedocs.build/projects/models/en/322/project/ssp.html#message_ix_models.project.ssp.data.SSPOriginal) and [here](https://iiasa-energy-program-message-ix--322.com.readthedocs.build/projects/models/en/322/project/ssp.html#message_ix_models.project.ssp.data.SSPUpdate).

So the user will need to (a) click on the links, (b) go to those pages, (c) read them, and (d) understand the instructions on how to register and access to data. Those websites are not part of this repository/package. So if (c) and (d) are not clear, we can't do anything about it by changing the code in this package. The fix would be for those sites to provide a better user experience.

> But I'm wondering whether the files that we have up in the repo, are "fuzzed" data or not. Because, honestly, I can't tell if those that we have publicly available on the repo are real or not. Does anyone know?

The files in [the message_ix_models/data/ssp/ directory](https://github.com/iiasa/message-ix-models/tree/2619da07c3e9581cedfc56e669ba285472437bfc/message_ix_models/data/ssp) are, for the moment, the ‘real’/original files. By this https://github.com/iiasa/message-ix-models/pull/322#issuecomment-2765436009, I meant we should probably remove them. (That will not happen in this PR, rather in another, follow-up PR. We try to keep each one as atomic as possible.)

The files in [the message_ix_models/data/**test**/ssp/ directory](https://github.com/iiasa/message-ix-models/tree/2619da07c3e9581cedfc56e669ba285472437bfc/message_ix_models/data/test/ssp) are for testing only. They are the ones that are “fuzzed”.

PR #341, opened 2025-04-30
Adjust use of data paths in tests

khaeru:
This PR adjusts the values passed to the [path_fallback(…, where=…)](https://docs.messageix.org/projects/models/en/latest/api/util.html#message_ix_models.util.path_fallback) keyword argument. Specifically:
- The "tests" location is omitted by default for any module using this function. This prevents the code from using test (fuzzed, random, or subset) data in ‘production’ real use, which could inadvertently produce incorrect results.
- Fixtures (using pytest's [monkeypatch features](https://docs.pytest.org/en/stable/how-to/monkeypatch.html#monkeypatching)) are added which append " tests" to the WHERE global in each module. This allows *only* the decorated tests to access the test data, temporarily.

**More TBA**

## How to review

**TBA**

## PR checklist

- [ ] Update doc/whatsnew.


glatterf42 (approved):
Not sure why we have these timeouts at six hours on CI, our tests should not run that long. If this is about network connectivity, could we put some lower timeout parameter (and possibly flaky/skip marks) in place to avoid using so much time? Or is there some way to resolve the root cause of these?
Otherwise, looks good to me :)

khaeru:
Thanks for the review!

> Not sure why we have these timeouts at six hours on CI, our tests should not run that long. If this is about network connectivity, could we put some lower timeout parameter (and possibly flaky/skip marks) in place to avoid using so much time? Or is there some way to resolve the root cause of these?

Looking at the logs of e.g. [this run](https://github.com/iiasa/message-ix-models/actions/runs/14785047250/job/41511938839?pr=341):
- At the very end, the log shows the job getting up to:
  ```
  [gw1] [ 99%] PASSED message_ix_models/tests/model/test_bare.py::test_create_res[settings4-expected4] 
  ```
- This means that almost all of the tests have completed, but then it times out.
- Searching the logs for `material/test_build.py::test_build_bare_res` shows only one completed test, but there are two: https://github.com/iiasa/message-ix-models/blob/ea59cd93b77603f640d3aa370c9defc8a59362d0/message_ix_models/tests/model/material/test_build.py#L30-L39
- So this is the same one test that I have set to skip on Windows on this branch.

As **mitigation**, a "timeout" parameter is a good idea, I'll look into that, or possibly also skip/xfail on macOS.

As for a **fix**, we would indeed need to look into bottlenecks in the materials build process and see where caching, better vectorization, or other approaches could help. For the Windows cases we now skip, I observed they sometimes ran long (though not 6 hours; they didn't stall/time out) but then failed with Java heap space (memory) errors. So I guess there is some interaction between high memory use and long runtimes.

PR #350, opened 2025-05-14
Migrate .engage.workflow and related tools

khaeru:
Prompted by [this issue](https://iiasa-ece.slack.com/archives/CD0GBHHA4/p1747224289016179) (Slack link, non-public), this PR migrates two functions used in .transport.workflow.tax_emission():

- message_data.projects.engage.workflow.step_0()
- message_data.projects.navigate.workflow.tax_emission()

…along with the code on which these are dependent.

The files and Git history migrate from the message_data `dev` branch, which is the version that is currently used / known-to-work with message_ix_models.model.transport. Some of these files may have been changed on the message_data `ssp_dev` branch; those changes can later be merged into their new locations here.

Per [our usual migration approach](https://docs.messageix.org/projects/models/en/latest/howto/migrate.html), the files were selected with the following:
```py
BATCH = (
    dict(
        args=[
            # Filter some paths
            "--path=data/alps",
            "--path=data/buildings",
            "--path=data/tests/buildings",
            "--path=doc/model/buildings.rst",
            f"--path={S.base}/model/buildings/",
            f"--path-regex={S.base}/projects/navigate/(cli|report|workflow).py",
            f"--path={S.base}/projects/navigate/wp2",
            f"--path={S.base}/tests/model/test_buildings.py",
            f"--path={S.base}/tests/project/test_navigate.py",
            f"--path={S.base}/tests/project/navigate",
            f"--path={S.base}/tests/tools/test_utilities.py",
            f"--path={S.base}/tools/utilities/add_AFOLU_CO2_accounting.py",
            f"--path={S.base}/tools/utilities/add_CO2_emission_constraint.py",
            f"--path={S.base}/tools/utilities/add_FFI_CO2_accounting.py",
            f"--path={S.base}/tools/utilities/add_alternative_TCE_accounting.py",
            f"--path={S.base}/tools/utilities/add_budget.py",
            f"--path={S.base}/tools/utilities/add_emission_trajectory.py",
            f"--path={S.base}/tools/utilities/add_tax_emission.py",
            f"--path={S.base}/tools/utilities/remove_emission_bounds.py",
            f"--path={S.base}/tools/utilities/update_h2_blending.py",
            # Rename paths
            f"--path-rename=data:{T.base}/data",
            f"--path-rename=data/tests:{T.base}/data/test",
            "--path-rename=doc/model/buildings.rst:doc/model/buildings/index.rst",
[...]

PR #351, opened 2025-05-14
Manual adjustments for `.engage.workflow` migration

khaeru:
This branch makes additional adjustments to the branch for #350 after the semi-automated migration process.

These include:
- Adjust imports and data paths
- Add type hints.
- Format docstrings according to the code style.
- Use ScenarioInfo instead of get_nodes(), get_optimization_years().
- New tests for the added .tools submodules.

# How to review

- Read the added What's New entry for #350/#351 in [the preview build of the docs](https://iiasa-energy-program-message-ix--351.com.readthedocs.build/projects/models/en/351/whatsnew.html), and all the pages/functions it links to; ensure they clearly indicate the state of the code.
- Skim the diff.

## PR checklist

- [x] Update doc/whatsnew.


message_ix_models/tests/model/test_buildings.py
```diff
@@ -1,16 +1,26 @@
...
     report3,
 )
+from message_ix_models.util import package_data_path
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from message_ix import Scenario
+
+    from message_ix_models import Context
```
glatterf42, line 23: I like all the new type hints :)
I'm just wondering: what's the benefit of importing these only `if TYPE_CHECKING`? Does this save actual runtime for the tests?

glatterf42 (approved):
Thanks, looks good to me :)

message_ix_models/tests/model/test_buildings.py
khaeru: Not sure if it has any performance effect.

I usually follow the heuristic "things used _only_ for type checking in a module are imported in a TYPE_CHECKING block". The corollary is that the top-level `import`s contain _only_ things actually referenced/called in the module.

This habit has helped me a number of times to avoid issues with circular imports.

PR #362, opened 2025-06-12
Adjust to pandas 2.3.0

glatterf42:
In https://github.com/iiasa/message_data/pull/610, we noted that the fixes I proposed there should also be applied here as the files have recently been migrated. 

Let's use the occasion to also pin pandas 2.2.3 for the RTD requirements, which allows the docs to be build again (see https://github.com/iiasa/ixmp/issues/582; essentially, we'll have to use the pin until our other dependencies [gamsapi and xarray for the most part] accommodate pandas 2.3.0. For gamsapi in particular, we should reach out to them to ask for development updates/changelogs and request that they publish a new version with a fix soon).

## How to review

- Confirm that the ReadTheDocs preview build of the documentation is successful.

## PR checklist

- ~[ ] Update doc/whatsnew.~ Just internal fixes.


glatterf42:
Just noticing that we're testing v3.11.0 here, even though we have 3.11.1 out for ixmp and message_ix. Should we adjust our CI to cover both or just 3.11.1 or are we fine with 3.11.0?

glatterf42:
Not sure what's wrong with the codecov/project check. As the codecov bot itself notes: 

> All modified and coverable lines are covered by tests ✅

So this sounds like some kind of error on the Codecov side, maybe comparing lots of reports to few or so. Either way, I would argue this should not keep this PR from being merged.

Issue #379, opened 2025-07-23
Documenting MESSAGEix workflows for paper unpack COP28


yiyi1991:
This page consolidates a brief documentation of how the MESSAGE modeling work is applied in the upcoming stand-alone paper led by Alaa. 

- One-sentence introduction: 
it undertakes a systematic evaluation of the COP28 commitments. We examine **how different plausible interpretations** of the Article 28 commitments align with pathways consistent with limiting warming to 1.5°C.

- Rolling title: 
**Unpack COP28**

- Who is involved: 
11 authors; Alaa, Yiyi, Meas on the key workflows; Shotaro, Sid, and Oliver on providing the net-zero fossil application; and Matt, Volker, Carl

- Branch: 
https://github.com/iiasa/gst-bespoke-scenarios
Private for now. Public after the net-zero fossil paper led by Shotaro is released. 

- Details of the model application: 
We included 4 dimensions that reflect the range of plausible interpretations of what “net-zero” could mean under Article 28.
**Net zero of what types of emissions** – whether the net-zero goal applies to CO₂ emissions only or to all GHGs (functions here); 
**Scope of emissions covered** – whether the target includes energy-related emissions only or also accounts for emissions from industrial processes or other relevant sources;
**Net zero where** – in Annex I regions or non-Annex-I regions, or globally;
**Level of effort to phase out fossil fuels** – representing the year by which a net-zero primary fossil fuel can be achieved.
Finally by which year (2050, 2060, etc.). **Thus 172 scenarios in all.**
Scenario config see [here](https://github.com/iiasa/gst-bespoke-scenarios/blob/main/gst_bespoke_scenarios/data/Config_file_runs.csv). 

- Scenario identifier: 
e.g., model: net_zero_accounting, scenario: **ssp1.5c_v6.1_dac-ccs_sd_a2050CO2_na2070CO2_zf2070**
(starting scenario SSP2_low, version 6.1, energy supply and demand emissions only and industrial process emissions not included, reaching net zero CO2 in Annex I regions by 2050, reaching net zero CO2 in non-Annex-I regions by 2070, reaching net zero primary fossil fuels globally by 2070)

- Starting scenario:
Three options: NGFS_phase_5_1p5c_scenario, SSP2_low, SSP_LED_low(for net zero fossil)
[...]

PR #381, opened 2025-07-24
Add template for documentation issues

glatterf42:
Following up on today's discussion, I prepared an issue form/template based on the two existing documentation issues we already have:

- https://github.com/iiasa/message_data/issues/616 (private)
- https://github.com/iiasa/message-ix-models/issues/379

Please note that I still enable blank issues because that's what we currently have and the only goal of this PR is to help with the creation of documentation issues.

## How to review

I hope the file changes show a preview, otherwise, I'm not sure how we could check that everything renders properly. Maybe by merging this first to a fork and opening an issue there.

## PR checklist

- [ ] ~Add or expand tests;~ coverage checks both ✅
- ~[ ] Add, expand, or update documentation.~ Not changing code.
- ~[ ] Update doc/whatsnew.~ Maybe just announce on Slack/during next weekly?


glatterf42:
The two "failing" tests were at 99% successful and then were cancelled because of the timeout limit. Do we know why this happens? Our nightly CI also seems to run into this issue frequently of late.

glatterf42:
Sounds good, thanks :)

The tests do pass on both windows and ubuntu, so we can apply the skipif mark, though I don't exactly know the reason why it times out (but of course, we could say `reason="Times out for unknown reasons"`). I think we can include that in this PR or else in any "transport improvements" PR :)

PR #392, opened 2025-08-07
Transport improvements from 2025-W32

khaeru:
> #380 ← previous PR | following PR → #430

- [x] iiasa/message_data#613
- [x] Convert `load-factor-ldv.csv` to a subdirectory structure parallel to `pdt-cap/{label}.csv`.
- [x] #400

## How to review

@r-aneeque

- Check that the input data files now contain the correct data for the DIGSY-{BEST,WORST}-{C,S} scenarios. I tried to consolidate the many commits inserting/changing/reverting values in these files, for instance commits that used scenario labels like "SSP(2024).2" and "DIGSY-BEST" before the "DIGSY-BEST-S" etc. labels were added.
- Confirm that we can delete the branch for #415 and https://github.com/iiasa/message-ix-models/commits/transport/2025-w20-digsy/ **or**, if there are data changes there that need to be preserved, give a specific list of which values in which data flows.

@glatterf42

- Note that CI checks all pass.

## PR checklist

- [x] Update doc/whatsnew.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -6,6 +6,20 @@ Next release
 
 - Improve and extend :doc:`/material/index` (:pull:`418`).
   See :doc:`version 1.2.0 </material/v1.2.0>` for details.
+- Improve :mod:`.model.transport` (:pull:`392`).
+
+  - New properties :attr:`.transport.Config.code` and
+    :attr:`.transport.Config.label`.
+  - New base class :class:`.MultiFile`
+    to read input data from distinct files according to scenario label.
+  - New :class:`LoadFactorLDV`, replacing :py:`load_factor_ldv`
+    and allowing a distinct file according to scenario label.
+  - Add technology dimension to :data:`.elasticity_f`.
+  - Document :class:`.ScenarioCodeAnnotations`.
+
+- Extend :mod:`.digsy.structure.SCENARIO` (:pull:`392`).
+- :class:`.ItemSchemeEnumType` automatically creates aliases
+  for SDMX item scheme members with "-" in their IDs (:pull:`392`).
 - Expand :doc:`develop` documentation (:pull:`423`)
   with material previously in :mod:`message_data` docs.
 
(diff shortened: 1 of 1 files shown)


message_ix_models/data/transport/R12/ldv-activity.csv
```diff
@@ -18,20 +18,20 @@
...
-*,        R12_CHN,  2110, 15000  # TH high
+*,        R12_CHN,  2020, 16000  # TH high
+*,        R12_CHN,  2110, 16000  # TH high
 # *,      R12_CHN,  2020, 16000  # Previous value
 # *,      R12_CHN,  2110, 16000  # Previous value
 # *,      R12_EEU,  2020,  8962  # Previous value
 # *,      R12_EEU,  2110,  8962  # Previous value
-*,        R12_EEU,  2020,  9000  # TH low
-*,        R12_EEU,  2110,  9000  # TH low
+*,        R12_EEU,  2020, 12500  # TH low
```
khaeru, line 33: @r-aneeque did Takuya send you these new values for the low/high end of his estimated ranges?

If not, then the `# TH low` comment is not applicable to these values, and will confuse us and others about where the values actually come from. If the new values are from a different source, then we should comment out the existing lines and add new lines with appropriate comments.

Please let me know what the situation is, and I can make the changes.

khaeru:
> The tests for Python 3.9 and upstream version 3.8.0 are not passing, is that intentional?

Ah, no—I seem to have used some features that were not advertised as being added in 3.10 or later. Will adjust.

message_ix_models/data/transport/R12/ldv-activity.csv
r-aneeque: @khaeru No these are just experimental values (based on the literature) to see if it smoothes out CAP_NEW. So in this case, we should remove the "#TH low" tag. I would suggest not using this value as a default for now until we have a better understanding of CAP_NEW and then we can check with Takuya to update the values wholesale

Issue #431, opened 2025-09-24
Trackers for the BMT development (build Buildings)


yiyi1991:
# Where it sits
This issue tracks the development of the Buildings sector in MESSAGEix-BMT runs. 
A full run should be able to 
- build the Buildings sector in the baseline scenario, 
- implement demand-sector policies in policy scenarios, 
- and contribute to scenario inter-comparison with high-granularity building sector results. 

This issue can be closed once the first milestone is reached. 

Currently code sits on:
- Branches in message-ix-models:
  - [`iiasa:bmt_dev`](https://github.com/iiasa/message-ix-models/tree/bmt_dev).
- Branch in message_data:
  - [`iiasa:bmt_dev`](https://github.com/iiasa/message_data/tree/bmt_dev).

# What needs to be there

- [x] Collect all pieces, no matter outdated or one-time calibration codes, to make sure the scenario with high-granularity Buildings will be consistent with the most recent deliveries (SSP scenarios) of the MESSAGEix model to Scenario MIP, 25.08
- [x] Build the takeaway BM scenario, 25.09
- [x] Finish the following roadmap table

  Enduse | Input source | Static built | Vetted | Dynamic built
  -- | -- | -- | -- | --
  resid_cooling/heating/hot_water | STURM | ready |1st round |one iteration, not automatic
  comm_cooling/heating/hot_water | STURM | ready |1st round  |one iteration, not automatic
  resid_construction/demolition | STURM | ready |1st round |ready
  comm_construction/demolition | STURM | ready |1st round |ready
  resid_cook/app | ACCESS/GLANCE | ready | ready | ready
  afofio_therm/spec | calibration |   |   | ready
  non_comm | calibrated ACCESS | ready | ready | ready

- [x] Build reporting, legacy report ready
- [x] Prepare first merge (cover all Buildings functions except dynamic build as it comes later in the policy scenario steps), by 26.03

# Where is the next step
Once the ``build`` function of the Buildings sector is stable, the next steps are tracked in the parent issue of this issue.

yiyi1991:
With the merge of PR#486, this sub-issue is closed.

PR #433, opened 2025-09-28
Add BMT workflows and update functions to build Buildings module

yiyi1991:
This is a draft PR to track bmt development (issue #431, and its parent issue #483) for the milestone "build the Buildings sector on a baseline scenario". 

Key functions that may need code check
- message_ix_models.model.buildings.build.**build_B**
- message_ix_models.model.buildings.build.**prepare_data_B**

When the buildings module is run as part of the bmt workflow, it calls ``build_B``. It loads inputs (prices, STURM outputs, static demand) specified in `context.buildings` or from the config file `data/bmt/config.yaml`, calls `prepare_data_B` to derive parameter data from the base scenario and those inputs, and then applies the buildings spec to the scenario. 

The `prepare_data_B` function produces demand and technology parameters (input, output, capacity factor, etc.) for buildings commodities and the residual AFOFIO representation; it can also add materials linkage when`with_materials` is true.

The two functions should be able to refactor and apply to message_ix_models.model.buildings.**main** and message_ix_models.model.buildings.build.**prepare_data**, so that the same functions apply to previous projects (e.g., NAVIGATE). Another PR will be opened for such a possible refactor.

## How to review

- (Optional) Run ``mix-models bmt run --from="M cloned" "BM built"`` to see if it works on an R12 scenario. The function should work once the ``test_bmt.py`` passed, which includes ``test_build_B_runs_with_minimal_data`` and ``test_build_B_runs_with_materials``. 
- (Optional) Run ``mix-models bmt run --from="BM solved" "BM reported"`` to see if the report works. This calls the legacy report on message-data branch `bmt_dev`.
- The documentation makes sense.

## PR checklist
- [x] Update doc/whatsnew.


message_ix_models/model/material/build.py
```diff
@@ -273,3 +273,78 @@ def make_spec(regions: str, materials: str or None = SPEC_LIST) -> Spec:
         ) from None
 
     return s
+
+
+# same as build(), but context-based
```
khaeru, line 278: Here and in .buildings.build, I think it would be preferable to adjust the build() or main() function to take a Context object, and then decide what to do/not do based on settings attached to it, like a .material.Config or .buildings.Config instance.

This way we could hopefully avoid duplicating methods, and reduce the risk that 2+ different 'build' methods diverge from one another.

khaeru: (As always, happy to try to help make this change.)

message_ix_models/util/context.py
```diff
@@ -459,3 +461,8 @@ def write_debug_archive(self) -> None:
         Shorthand for :meth:`.Config.write_debug_archive`.
         """
         self.core.write_debug_archive()
+
+    def print_contents(self):
```
khaeru, line 465: Probably equivalent to:
```python
print(repr(context.asdict()))
```

yiyi1991: ah, thx

message_ix_models/model/material/build.py
khaeru: The key transport lines are:
https://github.com/iiasa/message-ix-models/blob/b51d8f08fd999c50bb0d3a8c4ef10d0d030c4fab/message_ix_models/model/transport/build.py#L647-L652
and:
https://github.com/iiasa/message-ix-models/blob/b51d8f08fd999c50bb0d3a8c4ef10d0d030c4fab/message_ix_models/model/transport/build.py#L529-L545

The latter seems complicated, but it is basically trying to cover for different CLI, testing, and other uses of the code, and also allow to selectively override one or a few Config settings. For a minimal implementation, you could try something like:
```py
if "material" in context:
    # Existing config instance, don't do anything
    log.info(f"Use existing materials config: {context.material!r}")
else:
    # There's no materials Config instance, create a new one
    context["material"] = material.Config(**kwargs)
```

message_ix_models/data/technology.yaml
```diff
@@ -2896,6 +2896,14 @@ sp_el_RC:
   input: [electr, final]
   output: [rc_spec, useful]
 
+sp_el_RC_RT:
+  name: sp_el_RC
```
macflo8:
```suggestion
  name: sp_el_RC_RT
```

message_ix_models/tests/model/test_bmt.py
```diff
@@ -0,0 +1,484 @@
...
+        {
+            "commodity": ["afofio_spec", "afofio_therm"],
+            "node": [nodes[0], nodes[0]],
+            "year": [2020, 2020],
+            "value": [0.1, 0.1],
+            "unit": ["GWa", "GWa"],
+            "time": ["year", "year"],
+            "level": ["useful", "useful"],
+        }
+    )
```
macflo8:
```suggestion
    common_dims = dict(
        node="R12_AFR", unit="GWa", time="year", level="useful", year=[2020, 2030]
    )
    prices = pd.DataFrame({"commodity": commodities})
    sturm_r = make_df("demand", **common_dims, commodity="resid_heat_electr", value=1.0)
    sturm_c = make_df("demand", **common_dims, commodity="comm_heat_electr", value=0.5)
    demand_static = make_df(
        "demand", **common_dims, commodity=["afofio_spec", "afofio_therm"], value=0
    ).assign(year=[2020, 2020])
```

macflo8 (commented):
This looks good to me overall!
I left a few minor suggestions. Mainly formatting improvements and small refactorings to make some things more concise.

If there is any function/module that I should particularly look at (e.g.: regarding modelling methodology), please tag me there.

khaeru:
Rebased on current `main`.

doc/bmt/index.rst
khaeru, line 1: Newly added documentation files need to be added to 1 of the existing `.. toctree::` directives somewhere, else they will not appear in the built docs.

I've done this in the commits added to the branch.

khaeru (approved):
Approved, pending checks passing. I've made a number of refinements on recent commits on this branch, will edit this comment to explain.

- Instead of a separate class `.bmt.config.BuildingsConfig`, expand `.buildings.config.Config` to carry the additional information needed.
  - This avoids the need for lots of new code to read files, set defaults, update, etc., since `.buildings.config.Config` already has those.
- `.buildings.build_B()` was a duplicate of `.buildings.main()`, with substantial overlap in code but a different name. Instead of such a duplicate:
  - Use the same function for both 'methods'.
  - Add a setting (`.buildings.Config.method`) to control which of the methods is used; this setting enters the function through an argument (`context`).
  - Use `if:`-blocks in the function to guard steps that happen in one method but not the other.
- Instead of calling `get_spec()` and then following with code to modify its return value, pass an argument that signals it to return something different, as desired.
- The organization of test modules should match the organization of the code being tested. So, for instance, the added `.model.buildings.build.prepare_data_B` should be tested in `.tests.model.test_buildings`. Tests for `.model.bmt` code should be in `.tests.model.test_bmt`.

message_ix_models/cli.py
```diff
@@ -172,6 +172,7 @@ def _log_threads(k: int, n: int):
     "message_ix_models.testing.cli",
     "message_ix_models.util.pooch",
     "message_ix_models.util.slurm",
+    "message_ix_models.model.bmt.cli",
```
khaeru, line 175: Note to self: in a follow-up PR, ensure this list is sorted.

message_ix_models/model/buildings/build.py
```diff
@@ -602,6 +638,249 @@ def prepare_data(
     return result
 
 
+def prepare_data_B(  # noqa: C901
```
khaeru, line 641: Note to self: complexity here is 27.

In a follow up PR:
- At minimum, add a comment and mention in pyproject.toml, similar to other complex methods.
- Try to reduce the complexity somewhat.

message_ix_models/model/bmt/utils.py
```diff
@@ -0,0 +1,288 @@
...
+            scenario, info, sturm_r, sturm_c, "bm_subtraction"
+        )
+
+    else:
+        raise ValueError(f"Unknown method: {method}")
+
+    return mat_demand
+
+
+def build_PM(context, scenario: "Scenario", **kwargs) -> "Scenario":
```
khaeru, line 245: This seems to be used directly as a workflow step in the function `.workflow.generate()`. A loose convention we have is that functions that are workflow steps should live in the same .workflow module where they are used, so this could be shifted.

message_ix_models/cli.py
khaeru: Resolved in #486.

message_ix_models/model/buildings/build.py
khaeru: > - At minimum, add a comment and mention in pyproject.toml, similar to other complex methods.

This is done in #486.

PR #451, opened 2025-11-07
Add reporting of resource extraction to IAMC structure

khaeru:
Per discussion 2025-11-03, this PR gives a proof-of-concept for implementation of one reporting 'table' (measure) using the Reporter class. See in particular the added module `.report.extraction` and [the function `.report.extraction.callback()`](https://github.com/iiasa/message-ix-models/blob/enh/report-extr/message_ix_models/report/extraction.py#L28).

Some things to observe about the implementation:
- Reporting for this measure **starts from** specific quantities that are already present in the Reporter, thanks to the default behaviour of message_ix and other code.
  - If other reporting tasks also use the same inputs, they can be supplied from memory/cache and not recomputed.
- For **aggregation**, lists of codes (in this case, for the `c`/`commodity` dimension) are derived from the same files (in this case, commodity.yaml) that describe the model structure.
  - This ensures a single point of reference/truth for both model-building and -reporting.
  - The function itself should require minimal adjustment in the future; its behaviour will follow changes to commodity.yaml (or use of different commodity code lists for particular model variants).
  - The same data structures from the same keys can be used in any other module.
- The `callback()` function:
  - First adds some tasks specific to this measure, to get all the right values ready in genno.Quantity format and 'full' dimensionality.
  - Then uses a utility class (IAMCConversion) to set up more tasks that 'finalize' the IAMC-structured data, perform some partial sums, do renaming, assign units, etc.

Some things to observe about the PR:
- The new tasks are added in a 5-line function. The function is in a submodule of `.report`; a 'flat'/wide arrangement of similar reporting code will make for easier maintenance and readability than 1 or a few files with >1000 lines.
- The new function/module is added to `.report.config._default_callbacks`.
  This ensures it is called every time `prepare_reporter()` is called (unless user code deliberately removes it).
- The variables `.report.util.NOT_IMPLEMENTED_{IAMC,MEASURE}` and the expectation of `.test_report.test_compare()` are adjusted to reflect the additions.

Future PRs could make similar changes.

## How to review

- Read the function linked above.
- Comment on whether documentation allows to (a) understand its behaviour and (b) follow it as a pattern for expanding reporting of other quantities.

## PR checklist

- [x] Update doc/whatsnew.

### DIFF
--- a/doc/api/report/index.rst
+++ b/doc/api/report/index.rst
@@ -139,12 +139,36 @@ API reference
 .. automodule:: message_ix_models.report
    :members:
 
+   General-purpose code:
+
    .. autosummary::
 
       Config
+      defaults
       prepare_reporter
       register
       report
+
+   The following submodules prepare reporting of specific measures or quantities:
+
+   .. autosummary::
+      :toctree: _autosummary
+      :template: autosummary-module.rst
+      :recursive:
+
+      extraction
+
+.. currentmodule:: message_ix_models.report.key
+
+Keys
+----
+
+.. automodule:: message_ix_models.report.key
+   :members:
+
+   Pre-set :class:`genno.Key` instances
+   and :class:`genno.Keys` in this module can be referenced across modules,
+   instead of hand-typing the same strings in multiple places.
 
 .. currentmodule:: message_ix_models.report.plot
 
@@ -167,17 +191,25 @@ Operators
 
    .. autosummary::
 
+      broadcast_wildcard
+      call
       codelist_to_groups
       compound_growth
       exogenous_data
       filter_ts
       from_url
+      get_commodity_groups
       get_ts
       gwp_factors
       make_output_path
       model_periods
+      node_glb
+      nodes_world_agg
       remove_ts
+      select_allow_empty
+      select_expand
       share_curtailment
+      zeros_like
 
    The following functions, defined elsewhere,
    are exposed through :mod:`.operator`
@@ -186,6 +218,8 @@ Operators
    .. autosummary::
 
       message_ix_models.util.add_par_data
+      message_ix_models.util.merge_data
+      message_ix_models.util.nodes_ex_world
 
    Other operators or genno-compatible functions are provided by:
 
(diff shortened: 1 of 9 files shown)


yiyi1991:
Thank you so much Paul! I was able to run `ctx.report.key = "all::iamc"` and generated one extraction variable output file. 

This is a good example to follow (e.g., for adding one variable category, and for adding building-related variables reporting). **Please feel free to merge this PR.** 

This is not the best place but let me list two things I immediately sensed difficulties if we want to reproduce more complex power/material reporting under the same framework. 

The first **byproduct** issue becomes ok after I took a scecond look. 
https://github.com/iiasa/message-ix-models/blob/dc816406cc009d4aeae0851b47f4cc9f0001c66a/message_ix_models/model/material/report/reporting.py#L1709-L1712
When dealing with Final Energy variables, after retrieving all `in` flows by `df = rep.get("in:nl-t-ya-m-c-l")` with the explicit naming `flow\|level\|commodity\|tech\|mode`, we filtered them by level `final` (as only energy commodities use that level). Then the byproduct gases need to be excluded or even treated as negative energy input in specific scopes (e.g., bf_gas are generated by blast furnace but mostly immediately used by basic oxygen furnace, similarly those coke oven gas, but they should be counted under emission accounting.). But we can work with an, e.g., `excluding_comm` list to the key `c` right?

The second one is **dealing with different levels mapping to the same variable**. 
https://github.com/iiasa/message-ix-models/blob/dc816406cc009d4aeae0851b47f4cc9f0001c66a/message_ix_models/data/material/reporting/fe.yaml#L379-L383
To handle this the material reporting simply uses pandas groupby after genno. After 10 minutes digging, I think I need additional help to get genno handle that flexibly too. 

In short, I think as a maintainer I get the idea of what to do with the commodity.yaml. My understanding is, the commodities with the same commodity names in gdx but different levels, they should be treated as different commodities here in this yaml? E.g., the commodity `steel` has 9 levels, including `product` and `new_scrap`, which go to different iamc variables. You will see more level names in the upcoming bilateral trade...
https://github.com/iiasa/message-ix-models/blob/dc816406cc009d4aeae0851b47f4cc9f0001c66a/message_ix_models/data/commodity.yaml#L61-L67

PR #453, opened 2025-11-11
Swap macos-13 → -15-intel runners in CI

khaeru:
Per actions/runner-images#13046:
- The `macos-13` runner images are deprecated and will become unavailable.
- A new image `macos-15-intel` is available.

The reason we were still using `macos-13` is that message-ix-models still supports ixmp/message_ix v3.8 and v3.9. These versions neither (a) bundled binary GAMS libraries for arm64 (such as needed on the `macos-latest` runner) nor (b) included iiasa/ixmp#532 that adapted JDBCBackend to locate these libraries within the local GAMS installation. Thus, tests for these versions can't be run on arm64 systems (like `macos-latest`) and had to run on `macos-13`, which is x86-64. The new `-intel` image makes it possible to run the tests.

## How to review

- Read the diff/

## PR checklist

- ~Add, expand, or update documentation.~ N/A; CI changes only.
- ~Update doc/whatsnew~ ditto.
- [x] (After approval, before merge) Rebase to drop the TEMPORARY commit.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -99,7 +99,7 @@ jobs:
         # Minimum version given in pyproject.toml + earlier version of Python
         # For this job only, the oldest version of Python supported by message-ix-models
         - { upstream: v3.8.0,  python: "3.10"  }  # Released 2024-01-12
-        - { upstream: v3.9.0,  python: "3.14" }  # 2024-06-04
+        - { upstream: v3.9.0,  python: "3.13" }  # 2024-06-04
         - { upstream: v3.10.0, python: "3.14" }  # 2025-02-21
         # Latest released version + latest released Python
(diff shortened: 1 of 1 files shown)


glatterf42 (approved):
Looks good to me for the most part. Just one question about this failure that we see only on `ubuntu-latest-py3.14-upstream >= 3.11.0`:

```
FAILED message_ix_models/tests/model/transport/test_report.py::test_simulated_iamc - AssertionError: assert {'Energy Serv...nger|LDV|BEV'} <= {'Transport|S...nger|LDV|BEV'}
  
  Extra items in the left set:
  'Final Energy|Transportation|Bus'
  'Energy Service|Transportation|Domestic Aviation'
```

Is that expected or should/could this be fixed by this PR, too?

PR #456, opened 2025-11-12
Improve IAMCConversion

### LINKED ISSUE #454: Handle quantities with 2+ `n`/`y` dims in IAMCConversion
@junukitashepard found a limitation of this code:
https://github.com/iiasa/message-ix-models/blob/cb47229541975f818a9b84611d89ad2380af6766/message_ix_models/report/__init__.py#L153-L162

The particular application was conversion of `out:nd-nl-t-ya-c` to 2 different quantities:
1. For "Gross Exports":
   - The `nl` (i.e. the exporting node/region) dimension should be mapped to the IAMC `region` dimension
   - The `nd` (the importing region) dimension should be collapsed into the labels for the IAMC `variable` dimension.
2. For "Gross Imports", it's the reverse:
   - The `nd` dimension should be mapped to the IAMC `region` dimension.
   - The `nl` dimension should be collapsed into the labels for the IAMC `variable` dimension.

The way the code currently sets the default existence and contents of the "rename" setting does not allow to do (2).

We identified a workaround, but **to resolve** on `main`:
- Add tests.
- Adjust the code.
  - In #451, similar (simplified) code was added for in the [.report.util.IAMCConversion](https://docs.messageix.org/projects/models/en/latest/api/report/index.html#message_ix_models.report.util.IAMCConversion) class. I think it has the same limitations. In fixing this issue, I'd prefer to drop the function in report/\_\_init__.py, or slim it down to simply chain to the class.
- Improve documentation.

khaeru:
Closes #454.

## How to review

**TBA**

## PR checklist

- [ ] Update doc/whatsnew.

### DIFF
--- a/message_ix_models/model/structure.py
+++ b/message_ix_models/model/structure.py
@@ -1,10 +1,11 @@
 import logging
 import re
-from collections import ChainMap
-from collections.abc import Mapping, MutableMapping
+from collections import ChainMap, defaultdict
+from collections.abc import Iterator, Mapping, MutableMapping
 from copy import copy
 from functools import cache
-from itertools import product
+from itertools import product, starmap
+from typing import Literal
 
 import click
 import pandas as pd
@@ -15,7 +16,7 @@ from sdmx.model.common import Code, Codelist
 from sdmx.model.v21 import Annotation
 
 from message_ix_models.util import load_package_data, package_data_path
-from message_ix_models.util.sdmx import as_codes
+from message_ix_models.util.sdmx import as_codes, leaf_ids
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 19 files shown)


message_ix_models/model/structure.py
```diff
@@ -217,6 +218,57 @@ def generate_set_elements(data: MutableMapping, name) -> None:
         data[name]["indexers"] = indexers
 
 
+@cache
+def get_technology_groups() -> dict[Literal["t"], dict[str, list[str]]]:
```
yiyi1991, line 222: Surprisingly found that input `c` + `l` and output `c` + `l` are not enough to describe a group of technologies in the current version. 

I tried to use it to get a group of "fossil power generation" technologies, which worked well. 
Then I moved on to "solar power generation" technologies. I found that the `solar_res_*`s in the current model version do not use input at all, and they are also not the only technologies that do not use input (also winds). 

Maybe it is not a bad idea to add dummy commodity input or real ones (to get proper Primary Energy|Solar).

khaeru: > I found that the `solar_res_*`s in the current model version do not use input at all, and they are also not the only technologies that do not use input (also winds).
> 
> Maybe it is not a bad idea to add dummy commodity input or real ones (to get proper Primary Energy|Solar).

That could be one way to do it.

Other options—not necessarily mutually exclusive:
1. Add function(s) that call `get_technology_groups()` and then **further process** the return values to add, modify, remove, combine, etc. So for instance, one could take one of the lists and filter only entries like `name.startswith("solar_res")` or similar.
2. Add extra "annotations" and to technology.yaml:
   ```yaml
   solar_res1:
     name: solar_res1
     description: Maximum solar electricity potential 1
     type: secondary
     sector: electricity
     output: [electr, secondary]
     groups: ["solar res", "foo", "bar"]
   ```
   …and improve get_technology_groups() to add such tech IDs to the named groups.
3. Add extra 'parent' technologies with 'children' to technology.yaml:
   ```yaml
   SOLAR RES:
     name: Group of all solar potential technologies
     child: [solar_res1, solar_res2, ...]
   ```
   (This is done in the transport technology.yaml, for instance [here](https://github.com/iiasa/message-ix-models/blob/c33fe60e47227e763009bec6401afe290bc66971/message_ix_models/data/transport/technology.yaml#L438-L441), but not yet in the 'main' one.)
   get_technology_groups() can then understand that "SOLAR RES" *should not* be added to the Scenario's `technology` set, but *should* be interpreted as/included in the list of groups.

In the case of (2), (3), and the "dummy" inputs, the information still lives in the YAML file; in (1) the information is produced by code. So I would prefer the former for that reason.

Can you give any example(s) (1 or more) of groups of technologies that need to be available? I can add them to the tests and then make sure that the code can produce them.

yiyi1991: Ah thanks. The 2 and 3 make quite much sense. We should avoid making gdx bigger indeed. 
I was basically going through the technology groups sitting in lhs/rhs here. 

https://github.com/iiasa/message_data/blob/dev/data/policies/default_config.yaml#L810-L881

Will come back with examples once I finish.

PR #488, opened 2026-03-17
Add HOWTO Develop reporting code

khaeru:
This PR adds a new HOWTO guide that covers current recommended practice in developing and using reporting codes.

## How to review

Look at the added page in the ReadTheDocs preview build of the documentation:
https://iiasa-energy-program-message-ix--488.com.readthedocs.build/projects/models/en/488/howto/report.html

## PR checklist

- [x] Update doc/whatsnew.

### DIFF
--- a/doc/api/report/index.rst
+++ b/doc/api/report/index.rst
@@ -81,6 +81,8 @@ The basic **design pattern** of :mod:`message_ix_models.report` is:
   (and some extensions specific to :mod:`message_ix_models`)
   to handle the different sections of the file.
 
+.. _report-features:
+
 Features
 ========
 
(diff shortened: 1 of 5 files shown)


khaeru:
> What I feel though is giving some concrete examples of the two paradigms might be helpful. Not fully abstract, but stylized examples covering the same module (ideally energy since its common to all message scenarios as a Rosetta stone) to give a clear concretization.

I agree this would be valuable—as I understand, essentially showing 3 ways to report a single thing, with equivalent output.

Since some of the methods require verbose code and files it would be a big task to develop such an example, even if it was only demo code and not functional.

Since I don't have any bandwidth to add that in the near term, I will merge this so it becomes visible in the latest docs, and make an issue that someone can pick up later.

PR #432, opened 2025-09-26
Add reduced basin functionality

Wegatriespython:
### Reduced basin filtering for local nexus development

Adds basin filtering to enable running the nexus module locally for development.
Reduced basin mode is not intended for production, basin reduction is not
physically realistic. However for development of scenarios, model capabilities or policy design, the reduced basin scenario needs to give a good proxy of the whole model, hence the granular control over its implementation.

Addresses #414. Builds on #405 (now merged).

## Design

Basin selection is two-step and compositional:

1. Automatic selection per region: either `first_k` (head N from CSV order)
   or `stress` (N basins sampled across the demand/supply ratio spectrum).
2. filter_list augmentation: if specific basins are named, they are added
   to the automatic set (union, not replacement). E.g. automatic gives
   {A1, B1, C1}, filter_list [B1, B2, B3] gives {A1, B1, B2, B3, C1}.

All regions are guaranteed coverage by the automatic step.

## New CLI options (on `mix-models water-ix nexus`)

```
--reduced-basin / --no-reduced-basin
                                Toggle for A/B testing between full and
                                reduced basin sets.
--num-basins INTEGER            Basins per region (default 3).
--basin-selection [first_k|stress]
                                Automatic selection method (default first_k).
                                stress ranks basins by demand/supply ratio.
--filter-list TEXT              Additional basins on top of automatic
                                selection (repeatable).
```

### Python API

```python
# Toggle for A/B comment one line to switch
ctx.reduced_basin = True

ctx.basin_selection = "stress"   # or "first_k"
ctx.num_basins = 3               # per region
# Additive on automatic set:
ctx.filter_list = ["10|SAS", "115|SAS"]
```

## Files changed

| File | What changed |
|---|---|
| `utils.py` | `filter_basins_by_region()` (two-step selection), `compute_basin_demand_ratio()`, `_select_by_stress()`, `_diversity_select()` |
| `cli.py` | New click options on `nexus_cli` |
| `build.py` | Applies `filter_basins_by_region()` in `map_basin()`, stores `context.valid_basins` |
| `demands.py`, `infrastructure.py`, `irrigation.py`, `water_supply.py` | Filter data to `context.valid_basins` |
| `conftest.py` | `setup_valid_basins()` helper for tests |
| `test_utils.py` | Tests for stress selection, demand ratio, filter_list additive behavior |

## How to review

- Core logic: `filter_basins_by_region()` in `utils.py` - verify the two-step
  selection and the additive union of filter_list.
[...]

PR #509, opened 2026-05-15
Collect water module settings into Config dataclass

Wegatriespython:
Introduce `message_ix_models.model.water.Config` and move water-module state from loose top-level `Context` attributes to `context.water`.

The CLI now initializes and updates this config object for water options such as `RCP`, `SDG`, `REL`, time slices, region type, and reduced-basin settings. Build, data, and utility code read the same settings through `Config.from_context(context)`, matching the configuration pattern already used by transport and reporting. Following recommendation at the end of #432 from @khaeru  

Also adds a fix for nested CLI option handling so an omitted subcommand `--ssp` does not clear a value already set on the parent `water-ix` command.

## How to review
Look at the diff, check if the change to `click.py` might have un-intended side effects for other workflows.

## PR checklist

- [x] Update doc/whatsnew.


message_ix_models/tests/util/test_click.py
```diff
@@ -64,6 +64,26 @@ def inner(context, regions):
     assert "ZMB" == result.output.strip()
 
 
+def test_scenario_param_preserves_outer_value(mix_models_cli):
```
khaeru: Adding type hints to the signature (even `-> None:` at the end of line) would allow mypy to check the contents.

message_ix_models/model/water/data/demands.py
```diff
@@ -178,12 +179,13 @@ def add_sectoral_demands(context: "Context") -> dict[str, pd.DataFrame]:
     results = {}
 
     # Reference to the water configuration
+    cfg = Config.from_context(context)
```
khaeru, line 182: One thing that's worth thinking about briefly: is it the duty of these individual functions, or of higher-level code, to ensure the Context instance is properly prepared with a Config instance?

If the latter, this code could simply do:
```py
cfg: "Config" = context.water
```
…which would raise an exception if parent/calling code had not set up the proper context.

I believe it should work fine either way, since the way `.water.Config.from_context()` is currently written ensures that 2nd, 3rd, etc. calls will always get the _same_ Config instance as the first (thus, the same settings). (It does differ slightly from [`.transport.Config.from_context()`](https://github.com/iiasa/message-ix-models/blob/583e26773172a9d2b1af36abbb4d26faf9ad4bd3/message_ix_models/model/transport/config.py#L314-L317), which always creates a new instance. But IMO this difference is fine, for now, as long as the method docstrings are clear.)

message_ix_models/model/water/config.py
```diff
@@ -0,0 +1,61 @@
...
+if TYPE_CHECKING:
+    from message_ix_models import Context
+
+
+@dataclass
+class Config(ConfigHelper):
+    """Settings for :mod:`message_ix_models.model.water`."""
+
+    #: Water build variant.
+    nexus_set: Literal["cooling", "nexus"] = "nexus"
```
khaeru, line 15: This is great for now. In some cases it might make sense, in the future, to add an Enum, which can also help to sanitize user input:

```py
class NexusSet(Enum):
    #: (can add verbose description here)
    cooling = auto()
    #: (ditto)
    nexus = auto()

class Config(ConfigHelper):
    nexus_set: NexusSet = NexusSet.nexus

    def __post_init__(self) -> None:
        if isinstance(self.nexus_set, str):
            self.nexus_set = NexusSet[self.nexus_set]
```

khaeru (approved):
So far this looks great! I guess IDE tools can now jump from cfg.x references in the various functions to the places where Config.x are defined.

I see the RTD build is failing, but that is because of bugs upstream in ixmp4/pyam-iamc that I will try separately to mitigate. Those mitigations might belong here, or in message_ix or ixmp; whatever the case I'll notify to either re-run or rebase once they are in place.

> Also adds a fix for nested CLI option handling so an omitted subcommand --ssp does not clear a value already set on the parent water-ix command.

I see how this fix works, but I am struggling a little bit to think about use cases. It points to possibilities like this:
```
mix-models foo            bar            …  # (1)
mix-models foo --ssp=SSP1 bar            …  # (2)
mix-models foo            bar --ssp=SSP1 …  # (3)
mix-models foo --ssp=SSP1 bar --ssp=SSP1 …  # (4)
mix-models foo --ssp=SSP1 bar --ssp=SSP2 …  # (5)
```
What would be a case of 'foo' and 'bar' such that (2) and (3) need to be distinguished? Where (4) should change behaviour vs (2) and (3)? Where (5) must be supported?

Wegatriespython:
So the issue is more the water cli takes in some redundancies for options, which would be good to clean up as well.  

Currently for the water-ix group, `--ssp` uses `store_context`, which (pre-fix) when passed with `water-ix --ssp=SSP1 nexus`  would result in `None` from the missing ssp option to the nexus sub-command resetting the parent's value. 

With the fix, 2 and 3 will produce SSP1 and 5 will deliver SSP2.  

> ```
> mix-models foo            bar            …  # (1)
> mix-models foo --ssp=SSP1 bar            …  # (2)
> mix-models foo            bar --ssp=SSP1 …  # (3)
> mix-models foo --ssp=SSP1 bar --ssp=SSP1 …  # (4)
> mix-models foo --ssp=SSP1 bar --ssp=SSP2 …  # (5)
> ```

message_ix_models/model/water/data/demands.py
Wegatriespython: I think making it stricter makes sense, but currently there are a fair bit of changes coming up on water and I would like to leave it more defensive, with a docstring update.

khaeru:
> So the issue is more the water cli takes in some redundancies for options, which would be good to clean up as well.
> 
> Currently for the water-ix group, `--ssp` uses `store_context`, which (pre-fix) when passed with `water-ix --ssp=SSP1 nexus` would result in `None` from the missing ssp option to the nexus sub-command resetting the parent's value.

OK, agreed that cleaning up those redundancies (in a future PR) would be the preferred approach, but again the current fix is good for the code as it is.

In general/the long run, my preference would be:
- It would be an error to give the same option twice, as in example (4) or (5).
- For simplicity, we stick with the default click behaviour that each option name is attached to 1 specific group or (sub)command in the hierarchy, and must appear between the group/command name and the following subcommand (if any). We clearly document for users where each option must occur.
- There may be cases where two options with similar meaning are needed, but then we use different strings to distinguish them. For example: `mix-models foo-project --regions=R12 prep-data --from-regions=R99` could mean "take data *from* R99 and prepare data *for* an R12 model".

adrivinca (approved):
Also had a look. great tidy up!
I think it can be merged if the test failures are solved (and not due to changes here)

khaeru:
> I see the RTD build is failing, but that is because of bugs upstream in ixmp4/pyam-iamc that I will try separately to mitigate. Those mitigations might belong here, or in message_ix or ixmp; whatever the case I'll notify to either re-run or rebase once they are in place.

Now hopefully done in #511. Please rebase on `main` and let me know if that fails to fix CI issues.

PR #486, opened 2026-03-11
Add T into the BMT workflows and update MACRO for BMT

yiyi1991:
This is a PR to include 

- transport build, 
- transport report, 
- add MACRO, 
- full report chain (M+T+legacy report including B), 
- and other series updates for project implementation. 

See the tracker of bmt development (issue https://github.com/iiasa/message-ix-models/issues/483, milestone for NGFS C1 submission).

This PR branch is based on…
- ~`transport/2026-w04`~ was merged in #471.
- ~`transport/2026-w16`~ was merged in #497.
- `main`.

## How to review

- Read the diff and note that the CI checks all pass.
- The documentation makes sense.

## PR checklist

- [x] Continuous integration checks all ✅
- [x] Add or expand tests; coverage checks both ✅ (https://github.com/iiasa/message-ix-models/pull/486#issuecomment-4593848081)
- [x] Add, expand, or update documentation.
- [x] Update doc/whatsnew.

PR #510, opened 2026-05-15
Remove endogenous split of disconnected water demands

Wegatriespython:
This PR retires the connected/unconnected distinction for urban and rural water demands. 

The logic for retirement is the following : 

{urban/rural}_disconnected demands do not have formal infrastructure, hence no electricity demand on the model. Hence by retiring— where all demand goes via connected (infra backed) channels—we overstate the electricity demand. However numerically checking the results there is no aggregate change (at R12) in the energy response from representing urban/rural demand as fully connected.  

Given this lack of endogenous model response, this distinction can be recovered in post-solve at reporting time by splitting the urban/rural demands by connection rates. 

## How to review

This PR is based on #509's changes and can be merged only after #509 is merged. 

## PR checklist

- [x] Update doc/whatsnew.


adrivinca (approved):
We discussed about this and agreed on the change, which is also small in terms of code.
I would approve merging. why do we get the codecov? can we bypass it?

message_ix_models/model/water/report.py
adrivinca, line 478: Maybe worth changing this with a weighed average? althugh we lack data like population at the basin level.
This line of code is not new, but while before there was probably a discrepancy in what the model saw (exact basin access rates) vs what reported (simple average for population reporting), now we only ahve this, which might be too simplified

adrivinca:
@khaeru can we please merge this and move on? we have 2-3 big and small water and impacts PRs that should move on asap. Thanks!

@Wegatriespython probably rebase and check the testfail again

khaeru:
> @khaeru can we please merge this and move on? we have 2-3 big and small water and impacts PRs that should move on asap. Thanks!

Howdy —I haven't looked into this until now. Generally (because there are many PRs) I wait until I see the 4 items on the PR checklist completed and all green checks; and/or I get a direct request.

For codecov/patch, it seems that the function `.model.water.report.get_rates_data()` is not run as part of the test suite, and possibly also the function `report()` in the same file. Two ways to address this:
1. Add a test. A minimal test is one line that just checks the function runs, and doesn't include any detailed assertion about its effects. If it's hard to set up the conditions for the function to run successfully, it's OK to create a test that runs the function and marks it XFAIL; in this case the lines run will still be counted as covered.
2. Add an exclusion (`# pragma: no cover` comment) to the function and record in an issue somewhere that this is to be addressed later.

As you note, a rebase is also needed, so please let me know (e.g. request review) when that + one of the 2 things above is done.

Wegatriespython:
@khaeru, used pragma : no cover for codecov here since we were planning a migration of the water reporting code to be more genno intensive in the future. I see this macos test fail and the code quality seems to be gated on some fork safety policy. Not sure. Otherwise this should be good to go. Its been rebased on main.

khaeru:
OK, thanks for the change and the rebase.

> I see this macos test fail

I think this is just flakiness, should go away on a re-run.

> and the code quality seems to be gated on some fork safety policy. Not sure.

Thanks for the pointer. It looks like a consequence of the version bump for GitHub's actions/checkout in #525. I didn't notice the implication because I skimmed the changelog too quickly when merging that. Because it's my error, I will make an adjustment on `main` and then rebase this for you.

Wegatriespython:
@khaeru Did the rebase, seems like the test workflow issue is still active and not progressing due to the fork safety policy from earlier.

khaeru:
> @khaeru Did the rebase, seems like the test workflow issue is still active and not progressing due to the fork safety policy from earlier.

After finishing with review/merge of #486 and #549, I'll now rebase; look to address the CI issue; and then review.

khaeru:
> I'll now rebase; look to address the CI issue;

The CI issue was hopefully addressed by #550.
I resolved the conflict in doc/whatsnew while rebasing.

> and then review.

Coming up next.

message_ix_models/model/water/report.py
```diff
@@ -951,7 +976,9 @@ def compute_cooling_technologies(
     return report_iam, cooling_rows
 
 
-def report(sc: Scenario, reg: str, ssp: str, sdgs: bool = False) -> None:
+def report(sc: Scenario, reg: str, ssp: str, sdgs: bool = False) -> None:  # noqa: C901
```
khaeru, line 985: When C901 is ignored, we add a TODO line and an entry in pyproject.toml:
https://github.com/iiasa/message-ix-models/blob/b65ba84eadbd56dcf66854d614ae62b846c8fa4c/pyproject.toml#L192-L198

khaeru:
> The CI issue was hopefully addressed by #550.

Too optimistic. It seems like the jobs for this branch cannot create caches (that's OK) but also cannot *access* any caches created on the `main` branch, because those don't exist yet: the Test workflow hasn't run on the main branch since #550 was merged. I'll try to address this.

khaeru (commented):
Overall seems good, since it's mostly deletions.

A few things to change before merging—I will do these since I'm aware time is tight:
- Some commit messages are >50 characters, which doesn't conform to the [code style](https://docs.messageix.org/en/latest/contributing.html#code-style).
- I would drop the commit "Exclude untested integration-only branches…" that sprinkles `# pragma: no cover` and excludes all of `get_rates_data()`. I see there is no test for this function. The best approach in this case (and not too slow) is to _add_ such a test, even if it's a minimal one and must be marked xfail for now.
- One comment inline.

khaeru:
> It seems like the jobs for this branch cannot create caches (that's OK) but also cannot _access_ any caches created on the `main` branch, because those don't exist yet: the Test workflow hasn't run on the main branch since #550 was merged. I'll try to address this.

Did that through #553.

> `get_rates_data()`. I see there is no test for this function. The best approach in this case (and not too slow) is to _add_ such a test, even if it's a minimal one and must be marked xfail for now.

I also added a test for `.water.report.report()`. I saw a comment on the function about plans to refactor it using genno, but also looked at #513 and #522 and it does not seem to yet happen in those PRs.

This may not be enough to make the codecov/patch check pass, but it covers some of the gap and is a placeholder for a complete test to come.

khaeru (approved):

khaeru:
Thanks for this improvement, @Wegatriespython!

PR #513, opened 2026-05-20
Refresh water module data sources 

Wegatriespython:
This PR refreshes the input data sources for the water module, drops R11 support, and legacy impact representation for cooling module in lieu of PR #479  

- Water Availability :  
   - New set of ISIMIP3b sources for qtot, qr, and eflow from CWatM across GCMS. 
   - Percentile reliability recalculated, 
- Water demands : 
   - Withdrawal data for SSPs from Khan 2022 source, for return uses SSP2 return ratio from previous data (added as ratio file) applied uniformly across SSPS. 
- Water access rates: 
   -  Uses 2 sources, `projections_people_UR_income_10_25.csv`: with urban/rural split, used
  for AFR, EEU, FSU, LAM, MEA, PAS, RCPA, SAS, WEU.
   - `projections_people_merge_countries_10_25.csv`: no split, used for
  NAM and CHN with the same rate assigned to both settings.
   - PAO had anomalous values with the new data, hence its is hardcoded with 0.99 rate. 
- Desalination : 
  - Only SSP1, SSP3, SSP5 available, use mapping SSP1->SSP2, SSP3->SSP4
  - Downscales country level data to basin using a template derived from previous projected and historical desalination files. 
  - Adds a utility file `basin_allocation.py`  

Additionally, old files in pre-processing have been removed. 
   

## How to review
Go through generator files, to see methodology and assumptions for data imputation. 

- [x] Run nexus module pre and after update. (SSP2)
- [x] Run nexus module for other SSPs 

## PR checklist

- [x] Update doc/whatsnew.


adrivinca (commented):
just a few points:

1. get_rates_data fallback misses treatment/recycling rates for SSP≠2 (report.py:425). When all_rates_{ssp}.csv doesn't exist (which is every baseline run, since demands.py only writes it inside the if cfg.SDG != "baseline"  branch), the fallback globs {ssp}_regional_*_rate_*.csv. For SSP1/3/4/5 that only finds the connection-rate files —
  the treatment/recycling rates exist only as ssp2_* files. The connected/unconnected reconstruction works, but the water-access/sanitation reporting (the other get_rates_data call at line 705) silently loses treatment rates for  non-SSP2 runs. The fallback should mirror demands.py's "rates always come from ssp2" convention.
 
  2. Sub-annual demands now break for SSP≠2: we know. Let's maybe keep track of this as next step

  3. Stray no_climate file: e-flow_5y_m_no_climate_R12.csv was modified (+217/−217) even though no_climate was removed from _RCPS and the generator doesn't produce it — the qtot/qr no_climate files were deleted. Looks like a leftover;
  any reasons for it?

Wegatriespython:
@adrivinca Addressed two issues : 

1) The SSP2 treatment/recycling logic is refactored into helper so it is consistently used for demands.py & reporting.py and the glob is fixed to use SSP for connection and treatment/recycling from the helper. 
2) The stray 'eflow_noclimate' file has been deleted 

Sub annual left for future PR

adrivinca (approved):
need rebase and wait for #510

Wegatriespython:
@adrivinca rebased

message_ix_models/model/water/cli.py
```diff
@@ -97,18 +97,18 @@ def water_ini(context: "Context", regions, time):
     # setting the time information in context
 
 
-_RCPS = ["no_climate", "6p0", "2p6", "7p0"]
+_RCPS = ["2p6", "7p0", "8p5"]
```
khaeru, line 100: This looks like it might be better as an `enum.Enum`.

I wouldn't ask for it in this PR, but please consider for future development.

message_ix_models/model/water/data/pre_processing/basin_allocation.py
```diff
@@ -0,0 +1,72 @@
+"""Shared basin-allocation helpers for water input refreshes.
+
+The committed water data uses MESSAGE basin labels such as ``120|MEA``. Some
```
khaeru, line 3: I don't think we have this mentioned explicitly in our [code style](https://docs.messageix.org/en/latest/contributing.html#code-style), but for multi-line docstrings and comments we would typically wrap at the maximum line length, which is 88 per our use of ruff/black.

message_ix_models/model/water/data/pre_processing/generate_access_rates.py
```diff
@@ -0,0 +1,207 @@
+"""Populate R12 x SSP x (urban, rural) drinking-water access rates.
+
+Reads ``Improved water services`` rows from two source files under
+``data/water/demands/drinking_water_access``:
+
+- ``projections_people_UR_income_10_25.csv``: with urban/rural split, used
```
khaeru: Use the reST role for files and paths ([docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-file)):
```
:file:`projections_people_UR_income_10_25.csv`
```

message_ix_models/model/water/data/pre_processing/generate_access_rates.py
```diff
@@ -0,0 +1,207 @@
...
+            )
+            table.to_csv(out)
+            print(
+                f"ssp{ssp} {setting}: wrote {out.name} "
+                f"(2020 mean={table.loc[2020].mean():.3f}, "
+                f"2090 mean={table.loc[2090].mean():.3f})"
+            )
+
+
+if __name__ == "__main__":
```
khaeru, line 206: In the future, beyond/instead of having these modules be directly invokable like this, it would be best to tie them into the overall `mix-models` CLI:
- Remove the `if __name__ == "__main__":` blocks.
- Add a parallel module `.water.data.pre_processing.cli` (or use pre_processing/\_\_init__.py).
- Either:
  - Create a `click.group(…)` and decorate each of these `main()` functions with `@group.command(…)`, or
  - Define small wrapper functions in the new `.cli` module.
  
  (in either case, you can pass `help=__doc__` to use the module docstring for CLI help text.)
- Add to the top-level `message_ix_models.cli`.

message_ix_models/model/water/data/pre_processing/generate_hydro_availability.py
```diff
@@ -0,0 +1,399 @@
...
+
+import numpy as np
+import pandas as pd
+
+from message_ix_models.util import package_data_path
+
+log = logging.getLogger(__name__)
+
+HYDRO_PREPROCESS_DIR = Path(
+    "/mnt/p/watxene/ISIMIP_postprocessed/data_for_vignesh/message_nexus_input_2026"
```
khaeru, line 63: If the script depends on certain local paths, that should be documented.

doc/whatsnew.rst
```diff
@@ -8,6 +8,17 @@ Next release
 - Add IAMC code list :class:`~.iamc.structure.CL_SCENARIO_DIAGNOSTIC` (:pull:`501`).
 - New module :ref:`tools-newclimate` (:pull:`499`).
 - Add :class:`.model.water.Config` to collect water module settings (:pull:`509`).
+- Refresh :mod:`.model.water` input data and pre-processing (:pull:`513`).
```
khaeru, line 11: Below there are top-level bullets like "Improve buildings", "Improve transport", "Improve water", with each having either (a) changes from a single PR, or (b) changes from multiple PRs listed all together or with particular PRs for each sub-bullet.  In some releases (where there are very extensive changes to a particular module), we use sub-headings.

I re-read the whole change log before release and tidy/order it, but in the future please combine with existing bullets/text.

message_ix_models/model/water/data/pre_processing/generate_hydro_availability.py
```diff
@@ -0,0 +1,399 @@
...
+                out[:, j * 12 + (m - 1)] = np.nan
+                continue
+            out[:, j * 12 + (m - 1)] = bin_block[:, col_mask].mean(axis=1)
+    return out
+
+
+# ---------- Output formatting ----------
+
+
+def _annual_columns() -> pd.Index:
```
khaeru: This only appears to be used in one place, right?

I could understand if the contents were a bit slow and `@functools.cache` was used to store a generated value, but as-is the additional function seems less useful than a clear comment where its return value is used.

message_ix_models/model/water/data/pre_processing/generate_sectoral_demands.py
```diff
@@ -0,0 +1,158 @@
...
+from __future__ import annotations
+
+from pathlib import Path
+
+import pandas as pd
+
+from message_ix_models.model.water.utils import KM3_TO_MCM
+from message_ix_models.util import package_data_path
+
+SOURCE_R12 = Path("/mnt/p/ene.model/NEST/water_demands/Khan2022/withdrawals/annual/R12")
```
khaeru, line 42: Ditto above re: hard-coded paths.

khaeru (changes requested):
I made several comments inline, but all of them are informational or could be picked up in your next PR(s).

The one change I would request:
 
- Please reword the commit messages to follow the [code style](https://docs.messageix.org/en/latest/contributing.html#code-style), particularly the maximum length of the subject line.
  
  Also some commit message bodies seem cryptic and verbose, so perhaps you could revise them while thinking of colleagues who may need to interpret these code changes. For instance:

  - I would avoid words like 'shipped' (we don't ship), 'landed', etc.
  - "earlier in this PR" will be hard to reconnect with its original context. You can say things like "Adjust to use data added in #513", which will be auto-linked and also allows a reader to find the PR themselves.
  - It's not clear why some commit messages end with "Refs #535". The connection between these commits (or this PR) and that one is not explained. Please either explain or remove.

General observations:
- The "Code quality" check is failing because of some internal mypy error. I saw the same error (with the same upstream package, 'pint') [here](https://github.com/transportenergy/database/pull/112) and fixed it by bumping mypy to a newer version.

  If the error recurs here after you push the revised commit messages, we can add another commit making the same change.
- For the revised data files/future revisions, please consider the split between data which is not available elsewhere, and data that can be regenerated from a deterministic combination of other data + code. For the latter, it is better to do it once on the user's machine and cache the result to avoid recomputing. This also can help keep the repo and its history smaller.

Wegatriespython:
@khaeru Thank you for the review and the inline comments. I have reviewed and made changes to ensure the commits are within the character limits and the commit bodies are more clear.  The erroneous reference to 535 was there because that was the id of the note which kept track of the water pr development, removed now. 

Also followed the ReST role for files, paths and made the ` _annual_columns` inline. 

For the generation/commit point, yes this time the source files themselves were mostly derived data. I think in a future update it would be good to make the whole chain transparent and generated where possible.

PR #518, opened 2026-06-02
Build BMT config in workflow generate() so transport reporting runs

Tyler-lc:
`generate()` set ssp/regions but never called `apply_bmt_config`, so `context.transport` was unbuilt on the `mix-models efc run` path and transport reporting silently skipped (the broad handler logged an empty reason).

- Call `apply_bmt_config(context)` in `generate()`, alongside the existing ssp/regions setup.
- Log the transport-skip warning with `exc_info=True`.

Verified `generate()` now builds `context.transport` (a `transport.config.Config` with `.spec`); a full build+solve was not run.

### DIFF
--- a/message_ix_models/project/efc/workflow.py
+++ b/message_ix_models/project/efc/workflow.py
@@ -76,7 +76,7 @@ def report(context: Context, scenario: message_ix.Scenario) -> message_ix.Scenar
         try:
             _run_transport_report(context, scenario)
         except Exception as e:
-            log.warning("Transport reporting skipped: %s", e)
+            log.warning("Transport reporting skipped: %s", e, exc_info=True)
     else:
         log.info("Transport reporting skipped (no transport pax demand).")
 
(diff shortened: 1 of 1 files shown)


khaeru:
@Tyler-lc @yiyi1991 in order for colleagues to make sense of this PR, now or in the future:
- Please use (and create if necessary) a label like `p:EFC`, not a prefix in the PR title. Please look at and follow the example of other PRs.
- Please set an assignee for the PR. Often this is the person who opened it.
- Please follow our [contributing guidelines re: the PR description template](https://docs.messageix.org/en/latest/contributing.html#a-use-the-template-to-write-the-pr-description). A simple way to do this is to edit the description to strike out all the items and give a reason:
  ```md

  ## PR checklist

  - ~Continuous integration checks all ✅~ N/A; will be handled as part of #54321.
  - ~Add or expand tests; coverage checks both ✅~ ditto
  - ~Add, expand, or update documentation.~ ditto
  - ~Update doc/whatsnew.~ ditto
  ```
- Please ask if existing/prior PRs and the documentation do not make the appropriate practice clear.

Tyler-lc:
@khaeru Thanks for the advice. I have added the tag. I will not put the project name in the title in the future and i'll use tags instead.

khaeru:
> @khaeru Thanks for the advice. I have added the tag. I will not put the project name in the title in the future and i'll use tags instead.

To clarify, per the title and PR checklist, you can and should **edit** them to make the changes I indicated.
- For the title, click the ✏️ icon to the right of the PR number.
- For the description, click the "..." at the top right of the description, then "Edit".

PR #519, opened 2026-06-03
Generalize Genno hydrogen reporting to multiple domains

Tyler-lc:
This PR adds additional reporting variable in line with the EFC project and adds several hydrogen producing and using technologies:
- h2_elec (removed)
- h2_elec_alk (added - producer)
- h2_elec_pem (added - producer)
- h2_elec_soec (added - producer)
- h2_pyro_elec (added - producer)
- h2_ct (added - user)

The reporting uses the Genno package and works alongside the legacy reporting.
The following types of reporting variable types have been added:
- in|* 
- out|*
Additionally, import and export of HVC has been also reported

The current changes also add hydrogen leakages to all hydrogen related technologies as a `emission_factor` and reports them accordingly. As there is no CO2 equivalent emission factor currently it has no impact on the solved scenarios. Reporting of leaked hydrogen can be removed entirely if necessary. 

The exporting system to Excel needs to be improved, but it is good enough to be merged.

## How to review
- [ ] Check that all requested reporting variables are present
- [ ] Check if reported variables are correct, compare against GDX file (I have done some spot checks and sanity checks already)

## PR Checklist
- [ ] Add documentation
- [ ] Continuous integration checks all

<details>

<summary>Details and tracking of modifications</summary>

This PR extends the Genno hydrogen reporting so it can report other variable types beyond hydrogen. The reporting entry point changes from `run_h2_reporting` to `run_sectoral_reporting`: the new method takes one or more domain names (e.g. `"hydrogen"`, `"power"`) and reads the corresponding `message-ix-models/data/<domain>/reporting/` folders for the YAML files that drive the reporting. `run_h2_reporting` is kept as a thin wrapper, so existing hydrogen-only callers are unaffected.

`project/efc/workflow.py` is updated in this PR to call the new method — `run_sectoral_reporting(domains=["hydrogen", "power"])`.

The Power slice ships as the first non-hydrogen domain: `data/power/reporting/power_flow_in.yaml` and `power_flow_out.yaml`.

With second commit, the methanol reporting in|* and out|* style are added. I am still missing the reporting about the trade and bunkering in general. Currently I am not sure whether the bunkering happens in Mt or GWA. I have not been able to identify yet through input/output coefficient of technologies what is the case here. 
I am also not sure what is the shipping demand unit currently. This also needs verification before shipping the reporting for these variables.
The reporter now needs to files: Material flows and Energy flows. Current implementation of the Genno reporting only handles one unit at the time as input unit. So Energy flows (mode: Fuel) are in GWa, material flows (mode: feedstock) are in Mt. There is a caveat here. That is that the production still happens in GWa regardless of the mode. So, actually, only final_material is in Mt inside of the model.

The workflow has also been updated to include these new reportings. 

Other feedstock reporting has been added now. this includes olefines, ammonia and ethanol. We may have to be careful about ethanol (it can come from eth_bio or bio_eth and they are kind of different).
Shipping is reported, but I'm still not sure what is the correct unit there. 

Added small slightly unrelated fix for hydrogen reporting. The reporter was looking for Primary LH2 to report hydrogen as shipping fuel. Not it has been corrected
</details>


khaeru:
Please:
- Re-read https://github.com/iiasa/message-ix-models/pull/518#issuecomment-4610892221, in particular regarding the PR template.
- Read, understand, and apply the [code style](https://docs.messageix.org/en/latest/contributing.html#code-style), in particular as regards commit messages.

Tyler-lc:
I'll try to stick to the standard as much as possible. I have set-up now a Ruff automatic lint that should do the check before committing files. I have not tested it yet though. I hope it works fine.

yiyi1991:
This PR was opened to track the development of the report for the project EFC (including genno-based Material report, Transport report, Hydrogen report, and the integration with the legacy report). In order to test and report near-term results, I rebased this branch to #486 to include more up-to-date transport and bmt workflow development. 

(Sorry if you were bugged to review as a code owner.)

As it is now more ahead of the original main project branch (`project/EFC_2026`), I suggest Luca and I switch to this branch as the center branch and close this PR.

PR #531, opened 2026-07-02
Documentation for CO2 removal technologies in MESSAGE

ywpratama:
This draft PR includes a proposal for the documentation of CO2 removal technologies in MESSAGE

## How to review
When reviewing this draft PR, please provide feedback on the following but not limited to:
- Parent location of the doc. At the moment, it is put under emissions, but `Energy` section might be an option too.
- Now CDR technologies has its own section within the `Emissions` page. Please suggest whether this is appropriate or should it be moved under `CO Emissions` section
- This doc is planned to include other novel CDR technologies beyond those that are already included in the scenario MIP scenarios setup, so not only BECCS and DACCS. Please comment if this is OK or should we just put BECCS and DACCS for this.

## PR checklist

- [ ] Update doc/whatsnew.

### DIFF
--- a/doc/global/emissions/message/index.rst
+++ b/doc/global/emissions/message/index.rst
@@ -95,6 +95,28 @@ CO2 mitigation options in the energy system include technology and fuel shifts; 
      - steam methane reforming with CCS
      - 90%
 
+CO2 Removal Technologies
+~~~~~~~~~~~~~~~~~~~~~~~~
+In MESSAGE, a range of technology-based CO2 removal approaches are represented. This representation includes bioenergy with carbon capture sequestration (BECCS), 
+direct air carbon capture and sequestration (DACCS), biochar for non-land-use applications, biomass burial, enhanced rock weathering (ERW), and ocean alkalinity enhancement (OAE).
+
+Bioenergy with CCS (BECCS)
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Direct Air Capture (DACCS)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Biochar for non-land-use applications
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Biomass Burial
+^^^^^^^^^^^^^^
+
+Enhanced Rock Weathering (ERW)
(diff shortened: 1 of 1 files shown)


doc/global/emissions/message/index.rst
```diff
@@ -95,6 +95,28 @@ CO2 mitigation options in the energy system include technology and fuel shifts;
      - steam methane reforming with CCS
      - 90%
```
volker-krey, line 97: I would suggest to add the "CO2 Removal Technologies" section, one level higher, i.e. directly below the "Emissions" heading in [doc/global/emissions](../../tree/main/doc/global/emissions). This would be consistent with the structure of the [IAMC Wiki](https://www.iamcdocumentation.eu/Model_Documentation_-_MESSAGE-GLOBIOM).

doc/global/emissions/message/index.rst
```diff
@@ -95,6 +95,28 @@ CO2 mitigation options in the energy system include technology and fuel shifts;
...
+Biochar for non-land-use applications
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Biomass Burial
+^^^^^^^^^^^^^^
+
+Enhanced Rock Weathering (ERW)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Ocean Alkalinity Enhancement (OAE)
```
volker-krey, line 118: Adding conventional CDR, so afforestation and reforestation, would be important so that the list is comprehensive.

volker-krey (changes requested):
Thanks, @ywpratama. My main suggestion is to move the section one level up, so that it appears directly under the "Emissions" section.

doc/global/emissions/message/index.rst
ywpratama: Thanks, Volker. As suggested, I just moved the section up one level. So, the CDR section is directly under emissions, together with Emission from energy (MESSAGE) and Emissions from land (GLOBIOM) sections.

doc/global/emissions/message/index.rst
ywpratama: I just added the placeholder for conventional CDR on land on the CDR methods section for now. Will need to coordinate with GLOBIOM team (Yazhen?) and Oliver for the documentation

PR #539, opened 2026-08-20
Add open-license energy balance sources

Wegatriespython:
This is a draft PR corresponding to the working branch for the osce-energy-resilience project. It adds support to fetch UNSD and EUROSTAT data to provide an alternative to IEA for energy balance data. 

Add tools/unsd.py and tools/eurostat.py: national energy balances from the UN Statistics Division and from Eurostat over SDMX, each as a registered ExoDataSource and as a load_data() function for consumers that use pandas and have no Computer. Both emit the same dimensionality and units as IEA_EWEB under their own key tags, so a consumer can swap providers without its n and y dimensions changing shape.

Add util.sdmx.fetch_data() for the part of the query that is specific to neither service.

Neither aggregation nor interpolation runs unless asked for, and absent observations are never filled. The UNSD balance carries nine fuel groups against the IEA product axis of sixty-nine codes, so it serves aggregate indicators and cannot drive a technology calibration;

## How to review
~na~

## PR checklist

~- [ ] Continuous integration checks all ✅~
~- [ ] Update doc/whatsnew.~

### DIFF
--- a/doc/api/data-sources.rst
+++ b/doc/api/data-sources.rst
@@ -11,6 +11,16 @@ Tools for specific data sources
 .. automodule:: message_ix_models.tools.cepii
    :members:
 
+.. _tools-eurostat:
+
+Eurostat (:mod:`.tools.eurostat`)
+=================================
+
+.. currentmodule:: message_ix_models.tools.eurostat
+
+.. automodule:: message_ix_models.tools.eurostat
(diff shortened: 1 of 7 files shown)


Wegatriespython:
Refiling because gh PR stacking is messy over forks.

PR #543, opened 2026-08-20
Add open-license energy balance sources and statistics

Wegatriespython:
Add `SDMXSource`, an `ExoDataSource` for data retrieved from SDMX web services, and two concrete sources using it: `.tools.unsd` (UN Statistics Division energy balances and commodity-level statistics) and `.tools.eurostat` (Eurostat complete energy balances). These are openly-licensed alternatives to `.IEA_EWEB` for the countries either covers.

Both sources return data with the dimensions and units of `IEA_EWEB (n, y, product, flow; TJ)`,Country labels are mapped to ISO alpha-3 with pycountry, plus explicit handling of Eurostat's three non-ISO codes (Greece, the United Kingdom, Kosovo). No alpha-3 code raises. Product and flow codes are kept as the providers own for now and not translated to IEA codes. 

Units. Balances are queried in TJ. The UNSD commodity data are in physical units and are converted to TJ with the conversion factor UNSD attaches to each observation, following UNSD's own balance methodology; observations with no conversion factor (capacities, stocks) are dropped.

In the code, `SDMXSource` does the retrieval (cached) and the conversion to a `genno` `Quantity` once, following `IEA_EWEB` and DemoSource;

## How to review
The intention is not to merge into main, until it is clear that these sources would be desirable to land there. For now it still would require a review. Github PR stacking seems to not split out #554 out of this, so please ignore the diff from those changes here. 

This PR is prepared to support WS1 activities for OSCE. Energy balance information is used for regional baseline assessment, whereas wider energy statistics would be used for single country model preparation. 

- Country mapping. Check that the alpha-3 codes for OSCE countries are what you expect.
- Units and conversion. `TestUNSD_ENERGY.test_reconcile` checks Kazakhstan 2020 final consumption by commodity, converted and summed to fuel groups to reproduce UNSD's own balance values for the same year.
- Codes please check provider codes have been faithfully retained. 

## PR checklist

- [x] Update doc/whatsnew.


Wegatriespython:
@adrivinca @mawais210 Could either of you have a look at this PR and review it ?

adrivinca:
I had a human look and also used AI for some checks

### 1. Country mapping 
Good:
Eurostat side is good: `EL -> GRC` and `XK -> XKX` are correct and necessary. Its coverage
gaps (37/57 OSCE States) are by design, since `.tools.unsd` covers the rest.

The balance flow is good too: 57/57 OSCE States map cleanly.

Minor thing worth a note in comments:
 `CL_AREA_NRG` uses composite codes that are not ISO numerics, so
`on_missing="raise"` gives a hard `RuntimeError` with no partial result:

| code | UNSD label | resembles |
|---|---|---|
| `251` | France incl. Monaco | `250` France |
| `382` | Italy and San Marino | `380` Italy |
| `757` | Switzerland-Liechtenstein | `756` Switzerland |
| `579` | Norway incl. Svalbard and Jan Mayen | `578` Norway |
| `412` | Kosovo | — |

Affects `FRA ITA CHE NOR MCO SMR VAT`; four have no Eurostat fallback (`geo=CH` is not even
valid in `NRG_BAL_C`). They sit one digit from the real codes, so a careless fix silently
folds Monaco into France. that could be made clear.


### 2. Units and conversion — one real problem
(please double check, this is AI-derived)

The arithmetic is correct and the fixture is faithful: all 18 rows match the live service
exactly, values and conversion factors. Excluding `0129`/`0220` (which repeat `0100`/`0200`
tonnage) is right.

But two of eight groups do not reconcile with UNSD's published balance:

```
        test_expected  unsd_published  pct_diff
B00_CL    332276.1423     332276.1423    0.0000
B01_CP     40949.3786      40949.3786    0.0000
B02_PO       119.4158        119.4158    0.0000
B03_OP    508430.6796     559617.6192    9.1468   <-- 
B04_NG    256043.7450     271157.9823    5.5740   <-- 
B05_BW      2218.1460       2218.1460    0.0000
B07_EL    252440.3952     252440.3952    0.0000
B08_HT    275947.8180     275947.8180    0.0000
```

The six matching values agree to 1e-16, so they were transcribed from the source; the other
two equal the code's own output, which is why the test passes.

Gas: commodity `3000` at `TRANSACTION=12` is 284 493.05 TJ gross and is the only gas commodity
at that transaction. Published is 271 157.9823 — an implied factor of **0.9531**.

`tools/unsd.py:24` sets `NATURAL_GAS_GCV_TO_NCV = 0.9` and applies it at line 93 to every
observation with commodity `3000`, with no country or year dependence and no cited source.
For KAZ 2020 that gives 284 493.05 x 0.9 = 256 043.745 against UNSD's own 271 157.9823.

Likely single cause for both: `TRANSACTION=12` may not equal `B24_NA` in the balance flow. That
equivalence is assumed but never stated.

**The test suite cannot catch this.** `test_reconcile` reads the committed fixture
`data/test/unsd/DF_UNDATA_ENERGY.csv` as its input (via `monkeypatch` on `fetch_data`),
converts it, and asserts equality with the `expected` dict written a few lines below in
`test_unsd.py`. Both sides are committed constants and nothing contacts UNSD, so if a value in
`expected` is wrong the test still passes. That is what happened for `B03_OP` and `B04_NG`.

### To reproduce

Quickest, no code — open and compare `B03_OP` / `B04_NG` against the `expected` dict in
`test_unsd.py`:

```
https://data.un.org/ws/rest/data/DF_UNData_EnergyBalance/398..B24_NA.HSO?startPeriod=2020&endPeriod=2020&format=csv
```

Full check — needs only `requests` and `pandas`, no `message_ix`:

### 3. Provider codes — good

Little other suggestion:
- `UNIT_MULT` is fetched via `attributes="o"` but never read. The maths works only because
  `CONVERSION_FACTOR` applies to the already-multiplied value — assert it, or say why not.

Also for the future PRs it would be good to have in-person guide-through with me and @mawais210 so that we get into the topic and save time on reviewing

Issue #540, opened 2026-08-20
A plain install of message-ix-models cannot be imported


Wegatriespython:
In a fresh virtual environment, `pip install message-ix-models` followed by `import message_ix_models` raises an error. 

```python
  File ".../message_ix_models/__init__.py", line 21, in <module>
    from message_ix_models.util._logging import setup as setup_logging
  File ".../message_ix_models/util/__init__.py", line 11, in <module>
    import message_ix
  File ".../message_ix/__init__.py", line 6, in <module>
    from ixmp import ModelError, config
  File ".../ixmp/__init__.py", line 5, in <module>
    from ixmp._config import config
  File ".../ixmp/_config.py", line 445, in <module>
    config = Config()
  File ".../ixmp/_config.py", line 244, in __init__
    self.values = self._ValuesClass()
  File ".../ixmp/_config.py", line 72, in _platform_default
    ixmp4_databases = ixmp4.conf.settings.storage_directory.joinpath("databases")
AttributeError: module 'ixmp4.conf.settings' has no attribute 'storage_directory'
```
This is related to iiasa/ixmp#642.

With the first failure worked around, the import fails again:

```python
  File ".../message_ix_models/__init__.py", line 45, in <module>
    Context()
  File ".../message_ix_models/util/context.py", line 118, in __init__
    cls = getattr(import_module(module_name), "Config")
  File ".../message_ix_models/report/__init__.py", line 23, in <module>
    from .plot import prepare_computer as add_plots
  File ".../message_ix_models/report/plot.py", line 17, in <module>
    import plotnine as p9
ModuleNotFoundError: No module named 'plotnine'
```
This is related to #323 

Normally as a user these issues can be fixed relatively easily. However for the OSCE project I wish to declare message-ix-models as a dependency, and allow stakeholders to use the corresponding package. I don't want them to finagle with their venv to get things working.

Wegatriespython:
@khaeru For `plotnine` It seems moving `plotnine` to dependencies would resolve the issue, or we can add suitable guards to make it lazy import. 

For the ixmp issue I am not so clear how to fix.

khaeru:
Thanks for these checks! We don't catch this in our CI because we only install a full(er) set of dependencies: https://github.com/iiasa/message-ix-models/blob/4f24a7e36d3ce099c238d0c6cbe9307bf7092e76/.github/workflows/pytest.yaml#L205-L211

[...]

Issue #323, opened 2025-03-31
`tools.costs` requires `plotnine` dependency


glatterf42:
As discovered in #322, when installing message-ix-models in a fresh virtual environment from PyPI and trying to run 

```python
from message_ix_models.tools.costs.config import Config
```

I receive a `ModuleNotFoundError: No module named 'plotnine'`.

<details>
<summary>Full traceback</summary>

```python
>>> from message_ix_models.tools.costs.config import Config
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/fridolin/.virtualenvs/fix-git-lfs-pointers/lib/python3.12/site-packages/message_ix_models/__init__.py", line 30, in <module>
    Context()
  File "/home/fridolin/.virtualenvs/fix-git-lfs-pointers/lib/python3.12/site-packages/message_ix_models/util/context.py", line 118, in __init__
    kwargs[key] = cls(**values)
                  ^^^^^^^^^^^^^
  File "<string>", line 3, in __init__
  File "/home/fridolin/.virtualenvs/fix-git-lfs-pointers/lib/python3.12/site-packages/message_ix_models/report/config.py", line 26, in _default_callbacks
    from message_ix_models.report import plot
  File "/home/fridolin/.virtualenvs/fix-git-lfs-pointers/lib/python3.12/site-packages/message_ix_models/report/plot.py", line 15, in <module>
    import plotnine as p9
ModuleNotFoundError: No module named 'plotnine'
```
</details>

For installation, I simply ran `pip install message-ix-models` because there is no dependency section for `tools.costs`, indicating that no special dependencies are required. 

Manually installing `plotnine` fixes the issue. Note, though, that this is only testing a part of `tools.costs`, so other parts of the module may required additional dependencies that I haven't discovered so far.

So to resolve this issue, I suggest to add a `tools.costs` (or maybe `tools-costs` or just `costs`) section to `pyproject.toml` and list `plotnine` in there. 

To completely resolve the issue of any additional dependencies, I suggest running the full demo file and noting what else comes up as missing. Running the test suite is not really an option, I think, since that requires the `tests` section, which already brings in `report` and `transport`, including `plotnine`, so with that, we would even miss the dependency responsible for this issue.

khaeru:
> So to resolve this issue, I suggest to add a `tools.costs` (or maybe `tools-costs` or just `costs`) section to `pyproject.toml` and list `plotnine` in there.

[...]

PR #546, opened 2026-08-21
Fix/import without plotnine

Wegatriespython:
This PR addresses the `plotnine` import failure from the `message-ix-models` half of #540, and addresses #323 as well. 

`message_ix_models/report/plot.py` needs plotnine at import time. This PR makes its eager imports lazy.  There are two such sites: 
the `add_plots` re-export in `report/__init__.py`, now resolved through a module `__getattr__`, and the third default callback in `report/config.py`, now a wrapper that imports `.report.plot` when the callback runs instead of when `Config` is constructed. 

`types.py` imported `ggplot` solely to annotate `PlotAddable.__radd__`, so that import moves under `TYPE_CHECKING`

## How to review

Read the diff, check CI passes. Try local plain install of this branch once `ixmp` is pointed at iiasa/ixmp#640

## PR checklist

~- [ ] Add or expand tests; coverage checks both ✅~ CI side change required if any. 
~- [ ] Add, expand, or update documentation.~
- [x] Update doc/whatsnew.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -4,6 +4,7 @@ What's new
 Next release
 ============
 
+- Fix import of :mod:`message_ix_models` without :mod:`plotnine` (:issue:`323`, :issue:`540`, :pull:`546`).
 - Add IAMC code list :class:`~.iamc.structure.CL_SCENARIO_DIAGNOSTIC` (:pull:`501`).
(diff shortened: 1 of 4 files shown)


Wegatriespython:
Refiling, with branch pushed to origin because CI is not running on PRs from forks.

PR #547, opened 2026-08-21
Fix/import without plotnine

### LINKED ISSUE #323: `tools.costs` requires `plotnine` dependency

Wegatriespython:
This PR addresses the `plotnine` import failure from the `message-ix-models` half of #540, and addresses #323 as well. 

`message_ix_models/report/plot.py` needs plotnine at import time. This PR makes its eager imports lazy.  There are two such sites: 
the `add_plots` re-export in `report/__init__.py`, now resolved through a module `__getattr__`, and the third default callback in `report/config.py`, now a wrapper that imports `.report.plot` when the callback runs instead of when `Config` is constructed. 

`types.py` imported `ggplot` solely to annotate `PlotAddable.__radd__`, so that import moves under `TYPE_CHECKING`

## How to review

- Read the diff.
- Check that CI passes.
- Install `message_ix_models` from this branch without optional dependencies, particularly plotnine; run `python -c "import message_ix_models"`; and observe no ImportError.
- Try local plain install of this branch once `ixmp` is ~pointed at~ installed from the branch associated iiasa/ixmp#640.

## PR checklist

- [x] Update doc/whatsnew.

### DIFF
--- a/message_ix_models/report/__init__.py
+++ b/message_ix_models/report/__init__.py
@@ -20,7 +20,11 @@ from message_ix_models.model.workflow import STAGE
 from message_ix_models.util._logging import mark_time, silence_log
 
 from .config import Config
-from .plot import prepare_computer as add_plots
+
+try:
+    from .plot import prepare_computer as add_plots
+except ImportError:  # pragma: no cover
+    pass  # Plotnine not installed
 
 if TYPE_CHECKING:
     from genno.core.key import KeyLike  # TODO Import from genno.types
--- a/message_ix_models/report/config.py
+++ b/message_ix_models/report/config.py
@@ -24,12 +24,19 @@ Callback = Callable[[ComputerT, "Context"], None]
 
 
 def _default_callbacks() -> list[Callback]:
-    from message_ix_models.report import plot
-
     from . import defaults, extraction
 
     # NB When updating this list, also update the docstring of Config.callback
-    return [defaults, extraction.callback, plot.callback]
+    result: list[Callback] = [defaults, extraction.callback]
+
+    try:
+        from message_ix_models.report import plot
+
+        result.append(plot.callback)
+    except ImportError:  # pragma: no cover
+        pass  # Plotnine is not installed
+
+    return result
 
 
 @dataclass
--- a/message_ix_models/types.py
+++ b/message_ix_models/types.py
@@ -8,15 +8,7 @@ import sdmx.model.common
 from genno.core.key import KeyLike  # TODO Import from genno.types, when possible
 from plotnine import ggplot
 
-try:
-    from genno.core.quantity import AnyQuantity
-except ImportError:  # genno <= 1.24
-    from genno.core.quantity import Quantity
-
-    AnyQuantity: type = Quantity  # type: ignore [no-redef]
-
 __all__ = [
-    "AnyQuantity",
     "KeyLike",
     "MaintainableArtefactArgs",
     "MutableParameterData",
--- a/message_ix_models/util/cache.py
+++ b/message_ix_models/util/cache.py
@@ -22,8 +22,7 @@ import ixmp
 import sdmx.model
 import xarray as xr
 from genno import Computer
-
-from message_ix_models.types import AnyQuantity
+from genno.core.quantity import AnyQuantity
 
 from ._dataclasses import asdict
 from .context import Context
@@ -38,13 +37,16 @@ COMPUTER = Computer()
 # Show genno how to hash function arguments seen in message_ix_models
 
 
-def _quantity(o: "AnyQuantity"):
+def _quantity(o: "AnyQuantity") -> tuple:
+    """Encode :class:`genno.Quantity` `o` for cache key construction."""
     return tuple(o.to_series().to_dict())
 
 
 try:
     genno.caching.Encoder.register(AnyQuantity)(_quantity)
-except TypeError:  # Python 3.10 or earlier
+except TypeError:
+    # In Python <= 3.10, functools.singledispatch() does not support typing.Union
+    # Register each of the Quantity classes separately
     from genno.core.attrseries import AttrSeries
     from genno.core.sparsedataarray import SparseDataArray
 
--- a/doc/api/report/index.rst
+++ b/doc/api/report/index.rst
@@ -138,8 +138,11 @@ API reference
 
 .. currentmodule:: message_ix_models.report
 
+
 .. automodule:: message_ix_models.report
    :members:
+
+   :mod:`.report` requires the ``report`` :ref:`optional dependencies <install-dependencies>`.
 
    General-purpose code:
 
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -1,16 +1,17 @@
 Installation
 ************
 
-.. note:: :mod:`message_ix_models` requires :mod:`message_ix` to run.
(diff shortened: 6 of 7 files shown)


khaeru:
Thanks for opening this.

As a general principle, module-level `__getattr__()` should be reserved for cases where there is no simpler option. A chief reason is it frustrates static code analysis tools like type checkers, which can't follow imports that are determined dynamically at runtime.

One example where we _have_ used it is [here](https://github.com/iiasa/ixmp/blob/f2f906ed61d4954020160def87b2cfb3ff3ed7e7/ixmp/__init__.py#L55-L63). In this case it's for backwards compatibility/to avoid breaking code that imports from the old location, but the new location _is_ static and can be type checked etc.

I will experiment and push some commits to the branch for you to study.

message_ix_models/report/__init__.py
```diff
@@ -20,7 +20,11 @@
 from message_ix_models.util._logging import mark_time, silence_log
 
 from .config import Config
-from .plot import prepare_computer as add_plots
+
+try:
```
khaeru, line 24: This follows the Pythonic principle of [“easier to ask forgiveness than permission (EAFP).”](https://docs.python.org/3/glossary.html#term-EAFP)

message_ix_models/util/cache.py
```diff
@@ -22,8 +22,7 @@
 import sdmx.model
 import xarray as xr
 from genno import Computer
-
-from message_ix_models.types import AnyQuantity
```
khaeru, line 26: This was a fault in the existing code. The `.types` module is intended *only* for type checking, so it should only appear in code like:
```py
if TYPE_CHECKING:
    from message_ix_models.types import ...
```
One part of the fix was to correct this mistake.

message_ix_models/util/cache.py
```diff
@@ -38,18 +37,12 @@
...
 
-def _quantity(o: "AnyQuantity"):
+def _quantity(o: "AnyQuantity") -> tuple:
+    """Encode :class:`genno.Quantity` `o` for cache key construction."""
     return tuple(o.to_series().to_dict())
 
 
-try:
-    genno.caching.Encoder.register(AnyQuantity)(_quantity)
-except TypeError:  # Python 3.10 or earlier
```
khaeru, line 47: ~This is a vestigial work-around for genno 1.24.z and earlier. Since these versions are no longer supported (per pyproject.toml), I removed it.~

See [`functools.singledispatch()` in the Python docs](https://docs.python.org/3/library/functools.html#functools.singledispatch). I expanded the comment to explain why we need this.

message_ix_models/types.py
```diff
@@ -8,15 +8,7 @@
 from genno.core.key import KeyLike  # TODO Import from genno.types, when possible
 from plotnine import ggplot
 
-try:
-    from genno.core.quantity import AnyQuantity
-except ImportError:  # genno <= 1.24
```
khaeru, line 13: ~Ditto above re: genno 1.24.~ This is a vestigial work-around for genno 1.24.z and earlier. Since these versions are no longer supported (per pyproject.toml), I removed it. I also checked that this AnyQuantity is not being imported anywhere; many places in the code have:
```py
if TYPE_CHECKING:
    from genno.types import AnyQuantity
```
…so they were not relying on this.

message_ix_models/report/config.py
```diff
@@ -24,12 +24,19 @@
...
 
     # NB When updating this list, also update the docstring of Config.callback
-    return [defaults, extraction.callback, plot.callback]
+    result: list[Callback] = [defaults, extraction.callback]
+
+    try:
+        from message_ix_models.report import plot
+
+        result.append(plot.callback)
+    except ImportError:  # pragma: no cover
```
khaeru, line 36: Ditto above re: EAFP.

message_ix_models/types.py
```diff
@@ -8,15 +8,7 @@
 from genno.core.key import KeyLike  # TODO Import from genno.types, when possible
 from plotnine import ggplot
```
khaeru, line 9: Because (as above) this module is intended _only_ for type checking, it will _always_ be imported within an `if TYPE_CHECKING:` block. Thus there need not and should not be `if TYPE_CHECKING:` within this file.

This follows usage in some more sophisticated packages. A stricter approach would use ``assert TYPE_CHECKING`` to ensure first-party and downstream imports do not violate this.

khaeru (commented):
Please have a look at the explanatory comments I left inline and give a :+1: or ask any questions.

Once that's done, I'll approve; rebase to drop the reverted commits + reversions; and merge.

message_ix_models/types.py
Wegatriespython: Seems like we are under using this file then. Related to this I guess. https://github.com/iiasa/message_ix/discussions/1008

message_ix_models/report/__init__.py
Wegatriespython: I see, looks cleaner. However this would not address the startup costs issue no ? At least for those who have plotnine installed

khaeru: Sorry—to be clear at https://github.com/iiasa/message-ix-models/issues/540#issuecomment-5356915865 I was referring to pyam-iamc and/or ixmp4 re: start-up time. As far as I know, plotnine is not slow to import.

khaeru (approved):

PR #288, opened 2025-01-28
Check out correct ref in "pytest" CI workflow

khaeru:
#281 shifted the main event trigger for the "pytest" CI workflow from pull_request to pull_request_target. However, it erroneously omitted to explicitly specify the ref to be checked out. This is necessary because actions/checkout (and in this repo also francisbilham11/action-cached-lfs-checkout) will default to checking out $GITHUB_REF; for the pull_request_target event, this is the *base* branch commit, unlike for pull_request, where it is the source branch commit.

As a consequence of this, when merging #259, the "pytest" workflow appeared to pass, but this was in fact running the tests on `main`, while the tests did *not* all pass on the PR branch. Thus when #279 #282 were updated with rebase, they picked up these failing tests.

This PR:
- Corrects the pytest workflow, in a similar way to iiasa/message-ix-models#913.
- Fixes the tests that actually were failing on #259.

And also:
- Updates the copyright year to 2025.
- Drops support for ixmp/message_ix 3.4.0 and 3.5.0, in line with our stated [Upstream version policy](https://docs.messageix.org/projects/models/en/latest/develop.html#upstream-version-policy).
- Reduces the verbosity for some transport tests. The previous setting produced very verbose logs (~1.2 GiB) for some of the above failing runs.

## How to review

- Note the CI checks all pass.
- After this, I will rebase the branch to drop the one "TEMPORARY" commit, and then merge.

## PR checklist

- [x] Continuous integration checks all ✅
- [x] Add or expand tests; coverage checks both ✅
- ~Add, expand, or update documentation.~ N/A
- [x] Update doc/whatsnew.

PR #550, opened 2026-09-01
Use first-party actions to cache Git LFS objects

khaeru:
In #510 we see that the current config for PRs from forks can not coexist with our config to limit Git LFS bandwidth usage/associated costs.

- We use a multi-job configuration to first (in a `check` job), see whether the rest of the workflow should run at all, and if not error. See #281 and #288.
- GitHub's first-party actions/checkout introduced, in v7, an input `allow-unsafe-pr-checkout: true` that must be used in certain situations. The default of `false` is meant to protect users who don't have a config like ours.
- We have used [francisbilham11/action-cached-lfs-checkout](https://github.com/francisbilham11/action-cached-lfs-checkout/blob/main/action.yml) to avoid using Git LFS bandwidth on every job, but…
- …this action does not allow to pass the `allow-unsafe-pr-checkout: true` along to actions/checkout.

I adjust by using the steps suggested per https://github.com/actions/checkout/issues/165#issuecomment-2776048200, which is in an issue linked from the README of action-cached-lfs-checkout. In this way, we use actions/checkout directly, and can pass it the needed input.

Also:
- Mark some tests that are failing frequently because of network timeouts (perhaps rate limiting?) connecting to Zenodo.

## How to review

View [this run](https://github.com/iiasa/message-ix-models/actions/runs/33507112835) of the workflow on the "pull_request" trigger. This run was started because of the TEMPORARY commit on the branch; that commit will be dropped before merging.

## PR checklist

- ~Add, expand, or update documentation.~ N/A; CI changes only
- ~Update doc/whatsnew.~ ditto

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -257,7 +257,7 @@ jobs:
         mix-models config show
         # Symlink message-static-data into local data path
         mkdir -p ${{ github.workspace }}/local-data
-        cp -rsv $(realpath message-static-data)/* local-data/
+        cp -frsv $(realpath message-static-data)/* local-data/ || true
 
     - name: Run test suite using pytest
       run: |
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -37,7 +37,7 @@ jobs:
           || contains(github.event.pull_request.labels.*.name, env.label)
         )
       run: |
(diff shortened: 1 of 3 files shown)


khaeru:
Per the run linked above under "How to review", the CI is now passing with the changes on this branch. I'll rebase to drop the TEMPORARY commit and then merge without waiting for CI to re-run.

